PR #2, opened 2019-03-28
Add base Continuous Integration pipeline

jmaupetit:
## Purpose

We need a way to automate Docker production image build, validation & publication.

## Proposal

Add base CircleCI workflow to:

- [x] build production & development Docker images
- [x] check the production image
- [x] publish the production image to DockerHub upon git tags or PR merges

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,244 @@
+# Docker compose 3.4 file format requires a recent release of docker-compose
+upgrade-docker-compose: &upgrade-docker-compose
+  run:
+    name: Upgrade Docker Compose
+    command: |
+      curl -L "https://github.com/docker/compose/releases/download/1.23.2/docker-compose-$(uname -s)-$(uname -m)" > ~/docker-compose
+      chmod +x ~/docker-compose
+      sudo mv ~/docker-compose /usr/local/bin/docker-compose
+      docker --version
+      docker-compose --version
+
+version: 2
+jobs:
+  # Git jobs
+  # Check that the git history is clean and complies with our expectations
+  lint-git:
+    machine: true
+    working_directory: ~/fun
+    steps:
+      - checkout
+      # Make sure the changes don't add a "print" statement to the code base.
+      # We should exclude the ".circleci" folder from the search as the very command that checks
+      # the absence of "print" is including a "print(" itself.
+      - run:
+          name: enforce absence of print statements in code
+          command: |
+            ! git diff origin/master..HEAD -- . ':(exclude).circleci' | grep "print("
+      - run:
+          name: Check absence of fixup commits
+          command: |
+            ! git log | grep 'fixup!'
+      - run:
+          name: Install gitlint
+          command: |
+            pip install requests gitlint
+      - run:
+          name: lint commit messages added to master
+          command: |
+            gitlint --commits origin/master..HEAD
+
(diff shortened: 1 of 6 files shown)


jmaupetit:
Friendly reminder: this PR is ready for review :stuck_out_tongue: @sampaccoud // @lunika

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
...
+          requires:
+            - build
+          filters:
+            tags:
+              only: /.*/
+
+      # DockerHub publication.
+      #
+      # Publish docker images only if all build, lint and test jobs succeed and
+      # it has been tagged with a tag starting with the letter v
```
sampaccoud: ... or is on the master branch.

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
...
+            sudo sysctl -w vm.max_map_count=262144
+            sudo sysctl vm.max_map_count
+      # Run back-end (Django) check
+      #
+      # Nota bene: to run the django app, we need to ensure that both PostgreSQL
+      # and ElasticSearch services are up and ready. To achieve this, we wrap
+      # commands execution with dockerize, a tiny tool installed in the CI
+      # image. In our case, dockerize will wait up to one minute that both the
+      # database and elastisearch containers opened their expected tcp port
+      # (5432 and 9200 resp.).
```
lunika: All this comment is I think wrong. You are not using `dockerise` with `BUILD_TARGET` set to `production`. And we don;t have a `CI` image.

.circleci/config.yml
```diff
@@ -0,0 +1,240 @@
...
+            docker images | grep -E "^fundocker/funmooc\s*(${DOCKER_TAG}.*|latest|master)"
+      # Publish images to DockerHub
+      #
+      # Nota bene: logged user (see "Login to DockerHub" step) must have write
+      # permission for the project's repository; this also implies that the
+      # DockerHub repository already exists.
+      - run:
+          name: Publish images
+          command: |
+            DOCKER_TAG=$([[ -z "$CIRCLE_TAG" ]] && echo $CIRCLE_BRANCH || echo ${CIRCLE_TAG} | sed 's/^v//')
```
lunika: Did you try to put this in the task's `environment` section ? If it's possible to compute the value in one place I think we should do it.

env.d/production
```diff
@@ -0,0 +1,27 @@
+# Nota bene
```
lunika: It's weird to name this file `production` and only use it in the CI. I know it's to fit with what is done in the `docker-compose` file, but it's confusing IMO.

.circleci/config.yml
sampaccoud: I nearly commented the same but in fact I now think he is doing what this comment says, only via the make file.

env.d/production
jmaupetit: I know, this is why I wrote this preamble... I have no better option.

.circleci/config.yml
jmaupetit: In fact, @lunika is right: this is why I have added a `sleep 10` command :sweat_smile:

.circleci/config.yml
jmaupetit: I've defined environment variables. Let's see how it goes.

PR #3, opened 2019-04-03
🐛(settings) fix settings following changes on Richie master

sampaccoud:
## Purpose

Some settings have changed on Richie master that are breaking this repository (breaking changes are not taken care of as we are not yet in v1.0.0).

## Proposal

Apply the same changes as what was done in Richie:

- the `FUN_PLUGINS_GROUP` setting was renamed to `RICHIE_PLUGINS_GROUP`
- the `single-column.html` template was renamed to `single_column.html`
- the `fullwidth.html` template is kept only as a base template but can't be used anymore to build pages,
- the `create_demo_site` does not do its database partial cleaning anymore. We now rely on the `flush` Django command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ stop: ## stop the development server
 
 # == Django tasks
 demo-site:  ## create a demo site
+	@$(MANAGE) flush
 	@$(MANAGE) create_demo_site
 	@${MAKE} search-index;
 .PHONY: demo-site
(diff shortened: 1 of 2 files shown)


sampaccoud:
@jmaupetit Can't merge without CI anyway. Should I bypass it or wait for your other PR to be merged?

jmaupetit:
I think you need to bypass it since the CI in #2 is blocked by this anyway.

PR #5, opened 2019-04-10
Serve static and media files with AWS

sampaccoud:
### Purpose

We want to serve our static and media files with AWS S3 and CloudFront.

### Proposal

I basically copy/pasted what was done on [Marsha](https://github.com/openfun/marsha) and added a similar backend for Media files.

Some dependencies were missing after we have removed them from Richie master, so I proposed a commit to fix this as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # -- Docker
-UID              = $(shell id -u)
-GID              = $(shell id -g)
-COMPOSE          = UID="$(UID):$(GID)" docker-compose
+DOCKER_UID       = $(shell id -u)
+DOCKER_GID       = $(shell id -g)
+COMPOSE          = DOCKER_USER="$(DOCKER_UID):$(DOCKER_GID)" docker-compose
 COMPOSE_RUN      = $(COMPOSE) run --rm
 COMPOSE_RUN_APP  = $(COMPOSE_RUN) app
 COMPOSE_EXEC     = $(COMPOSE) exec
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ endif
 # -- Rules
 default: help
 
-bootstrap: data/media/.keep data/static/.keep build-front build run migrate  ## install development dependencies
+bootstrap: env.d/aws data/media/.keep data/static/.keep build-front build run migrate ## install development dependencies
 .PHONY: bootstrap
 
 # == Docker
@@ -84,22 +84,26 @@ watch-sass: ## watch changes in Sass files
 	@$(YARN) watch-sass
 .PHONY: watch-sass
 
-# == Django tasks
-check:  ## perform django checks
+#== AWS/Terraform
+env.d/aws:
+	cp env.d/aws.dist env.d/aws
+
+# == Django
+check: ## perform django checks
 	@$(MANAGE) check
 .PHONY: check
 
-demo-site:  ## create a demo site
+demo-site: ## create a demo site
 	@$(MANAGE) flush
 	@$(MANAGE) create_demo_site
 	@${MAKE} search-index;
 .PHONY: demo-site
 
-migrate:  ## perform database migrations
+migrate: ## perform database migrations
 	@$(MANAGE) migrate
 .PHONY: migrate
 
-search-index:  ## (re)generate the Elasticsearch index
+search-index: ## (re)generate the Elasticsearch index
 	@$(MANAGE) bootstrap_elasticsearch
 .PHONY: search-index
 
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,10 @@ check: ## perform django checks
 	@$(MANAGE) check
 .PHONY: check
 
+collectstatic:  ## collect static files to /data/static
+	@$(MANAGE) collectstatic
+.PHONY: collectstatic
+
 demo-site: ## create a demo site
 	@$(MANAGE) flush
 	@$(MANAGE) create_demo_site
--- a/README.md
+++ b/README.md
@@ -30,6 +30,11 @@ migration-related issues.
 Once the bootstrap phase is finished, you should be able to view the site at
 [localhost:8070](http://localhost:8070)
 
+> If you've just bootstrapped this project, you are probably planning to use AWS
+> to store and distribute your media and static files. Luckily for you, we've
+> cooked `terraform` scripts and a documentation for you! Read more about it:
+> [docs/aws.md](./docs/aws.md)
+
 ## Usage
 
 ### Managing services
--- a/bin/manage
+++ b/bin/manage
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# Run Django management command using docker-compose configuration (for
+# development only)
+DOCKER_USER="$(id -u):$(id -g)" \
+    docker-compose run --rm app \
+        dockerize -wait tcp://db:5432 -timeout 60s \
+		    python manage.py "$@"
--- a/bin/state
+++ b/bin/state
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# Run Terraform commands on the create_state_bucket project in the Hashicorp
+# docker container, passing our environment variables
+UID="$(id -u):$(id -g)" \
+    docker-compose run --rm \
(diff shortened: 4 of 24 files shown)


src/backend/funmooc/settings.py
```diff
@@ -609,13 +610,36 @@ class Production(Base):
...
+    AWS_ACCESS_KEY_ID = values.SecretValue()
+    AWS_SECRET_ACCESS_KEY = values.SecretValue()
+
+    AWS_S3_OBJECT_PARAMETERS = {
+        "Expires": "Thu, 31 Dec 2099 20:00:00 GMT",
+        "CacheControl": "max-age=94608000",
+    }
+    AWS_S3_REGION_NAME = values.Value("eu-west-1")
+
+    AWS_STATIC_BUCKET_NAME = values.Value("production-fun-mooc-static")
```
sampaccoud: @jmaupetit maybe it's better to use `production-funmooc-static` like you did on the project name...

src/aws/cloudfront.tf
```diff
@@ -0,0 +1,97 @@
...
+    s3_origin_config {
+      origin_access_identity = "${aws_cloudfront_origin_access_identity.fun_mooc_oai.cloudfront_access_identity_path}"
+    }
+  }
+
+  enabled         = true
+  is_ipv6_enabled = true
+
+  # Allow public access by default, served by static bucket
+  default_cache_behavior {
```
lunika: In the django's `storage.py` file, for static configuration you defined:
- `location = settings.STATIC_URL`

So I think you should you the `path_pattern` setting and set it to `/static/*`, WDYT ?

lunika: Or maybe create an other `ordered_cache_behavior` dedicated to `static` files ?

src/aws/s3.tf
```diff
@@ -0,0 +1,118 @@
+# Create S3 Bucket for static files
```
lunika: I think everything can be done in only one S3 bucket, you have a different location storage in your config so `media` and `static` will not be mixed but in a different folder.

src/aws/cloudfront.tf
sampaccoud: the static is covered by default behavior. Otherwise what would the default behavior be?

lunika: is it mandatory to have one ? Does it make sense to serve something else than `media` and `static` ?

src/aws/s3.tf
sampaccoud: I usually keep static and media in different buckets because they have a different life cycle. Media files should be versioned and backed-up whereas static files are committed in your repo so we don't care about them. WDYT?

lunika: :ok_hand:

src/aws/cloudfront.tf
sampaccoud: Yep:
> default_cache_behavior (Required) - The default cache behavior for this distribution (maximum one).

lunika: ok then keep the conf as it.

sampaccoud: What will happen is that you will hit the static bucket for anything else and get a 404... If we don't add anything else to the static bucket it should be ok?

bin/state
```diff
@@ -0,0 +1,14 @@
...
+# Run Terraform commands on the create_state_bucket project in the Hashicorp docker container,
+# passing our environment variables
+
+docker run --rm -it \
+    -u "$(id -u)" \
+    -v "${PWD}/src/aws/create_state_bucket:/app" \
+    -w "/app" \
+    --env-file ./env.d/aws \
+    "hashicorp/terraform:light" \
+    "$@"
```
jmaupetit: Why not adding a terraform service in the docker-compose file and add more make rules? or at least use a terraform docker-compose service here?

Makefile
```diff
@@ -41,6 +41,23 @@ stop: ## stop the development server
 	@$(COMPOSE) stop
 .PHONY: stop
 
+#== AWS tasks
+
+env.d/aws:
+	cp env.d/aws.dist env.d/aws
+
+deploy: ## Create or update the AWS infrastructure for fun-mooc
```
jmaupetit: deploy is confusing IMO

Makefile
```diff
@@ -41,6 +41,23 @@ stop: ## stop the development server
...
+	@bin/terraform apply
+.PHONY: deploy
+
+output: ## Display AWS infrastructure output configuration
+	@bin/terraform output
+.PHONY: output
+
+init: env.d/aws ## Initialize Terraform
```
jmaupetit: you should prefix all those rules as: `tf-xx` IMO

bin/manage
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# Run Terraform commands on the create_state_bucket project in the Hashicorp
+# docker container, passing our environment variables
```
lunika: This comment is wrong IMO

jmaupetit: Oops copy/pasta fail.

docs/aws.md
```diff
@@ -0,0 +1,123 @@
+# Statics and media files management with AWS
+
+We use AWS S3/CloudFront to store and distribute statics and media for the
```
sampaccoud:
```suggestion
We use AWS S3/CloudFront to store and distribute static and media files for the
```

docs/aws.md
```diff
@@ -0,0 +1,123 @@
...
+project in various environments. Buckets, distributions and IAM users creation
+is fully automated using [Terraform](https://www.terraform.io/).
+
+> ✋ If you plan to develop locally on the project, you don't have to configure
+> anything more. The following documentation targets operational users willing
+> to setup an AWS infrastructure to handle static and media files.
+
+## AWS Credentials
+
+To bootstrap the AWS infrastructure, you will need first to edit the `env.d/aws`
```
sampaccoud:
```suggestion
To bootstrap the AWS infrastructure, you will first need to edit the `env.d/aws`
```

docs/aws.md
```diff
@@ -0,0 +1,123 @@
...
+environment file with your AWS credentials and settings. You can start by
+copying the `env.d/aws.dist` template _via_ the `make aws/env.d` command. Now
+edit this new file with your AWS secrets:
+
```
sampaccoud: make it clearer that this credentials should have sufficient access rights to : create S3 buckets, KMS keys, dynamoDB, IAM users and CloudFront distributions?

docs/aws.md
```diff
@@ -0,0 +1,123 @@
...
+TF_VAR_aws_region=eu-west-1
+```
+
+## Setup a shared Terraform state
+
+> ✋ If the project already exists with a shared state, you can skip this section
+> and start fetching the state locally to create new workspaces.
+
+If the project doesn't exist at all, you will need to create an S3 bucket (and a
+DynamoDB lock table) to store your Terraform state file (and it's locking) by
```
sampaccoud:
```suggestion
DynamoDB lock table) to store your Terraform state file (and its locking) by
```

docs/aws.md
```diff
@@ -0,0 +1,123 @@
...
+
+If everything went smoothly, it's time to initialize your main terraform project
+using the shared state:
+
+```
+$ bin/terraform init
+```
+
+Now that your terraform project is initialized, you will be able to create S3
+buckets for media and statics in various environments (see the [project's
```
sampaccoud:
```suggestion
buckets for static and media files in various environments (see the [project's
```

docs/aws.md
```diff
@@ -0,0 +1,123 @@
...
+- `DJANGO_AWS_ACCESS_KEY_ID`
+- `DJANGO_AWS_SECRET_ACCESS_KEY`
+
+Corresponding values can be obtained using the following Terraform command:
+
+```
+$ bin/terraform output
+```
+
+Expected output should look like the following:
```
sampaccoud: `should` and `expected` seem redundant in the sentence:
```suggestion
The output should look like the following:
```

README.md
```diff
@@ -30,6 +30,11 @@ migration-related issues.
 Once the bootstrap phase is finished, you should be able to view the site at
```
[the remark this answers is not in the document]
jmaupetit: 😅

PR #6, opened 2019-04-17
⚡️(circleci) use docker executors to speed up execution

jmaupetit:
## Purpose

Following recent changes in Richie's Continuous Integration workflow, we've decided to backport optimization made in this project using Docker executors and preventing docker images manual caching to prefer Docker layer caching (DLC).

Note that for now this feature is not yet activated in our account leading to an increased DockerHub publication time (no cache build).

## Proposal

- [x] use docker executors
- [x] reduce the number of jobs


env.d/ci
```diff
@@ -0,0 +1,5 @@
+# Nota bene: this environment file is a superset of the development environment
+# to avoid repeting ourselves.
```
sampaccoud:
```suggestion
# to avoid repeating ourselves.
```

docker-compose.yml
```diff
@@ -42,6 +33,24 @@ services:
       - "db"
       - "elasticsearch"
 
+  app-ci:
```
sampaccoud: I'm not convinced that this should be in the docker-compose but :man_shrugging:

jmaupetit: We have no other option but to use docker-compose to test our production image in the CI. And instead of having multiple docker-compose files, I prefer defining a new dedicated service.

lunika: I agree with @sampaccoud

jmaupetit: I have no better options for now. I am open to suggestions.

PR #16, opened 2019-05-24
Upgrade to richie-1.0.0

jmaupetit:
## Purpose

While upgrading Richie to 1.0.0-beta.9, we forgot to upgrade our settings and urls definitions. In the meantime a new richie release came up, thus we've also switched to it!

## Proposal

- [x] use the new `RichieCoursesConfigurationMixin`
- [x] add `filer` urls
- [x] remove settings that are now defaults


jmaupetit:
@sampaccoud not sure that the filer URLs will play well with aws/s3?

PR #17, opened 2019-05-29
🐛(settings) fix CDN_DOMAIN value

jmaupetit:
## Purpose

The `CDN_DOMAIN` setting is empty (at least in `Staging` environment).

## Proposal

We need to set the `CDN_DOMAIN `setting to the same value as the `AWS_CLOUDFRONT_DOMAIN` setting. Let's instantiate it from the environment instead.

### DIFF
--- a/src/backend/funmooc/settings.py
+++ b/src/backend/funmooc/settings.py
@@ -378,7 +378,12 @@ class Production(Base):
 
     AWS_CLOUDFRONT_DOMAIN = values.Value()
 
-    CDN_DOMAIN = AWS_CLOUDFRONT_DOMAIN
+    @property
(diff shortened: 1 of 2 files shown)


src/backend/funmooc/settings.py
```diff
@@ -378,7 +378,9 @@ class Production(Base):
 
     AWS_CLOUDFRONT_DOMAIN = values.Value()
 
-    CDN_DOMAIN = AWS_CLOUDFRONT_DOMAIN
+    @property
+    def CDN_DOMAIN(self):
+        return self.AWS_CLOUDFRONT_DOMAIN
```
sampaccoud: needs a docstring?

PR #25, opened 2019-07-18
Add backend linters

jmaupetit:
## Purpose

Now that we produce backend code for this project, we need to automate its quality assessment.

## Proposal

- [x] add `Makefile` rules to run linters locally
- [x] automate quality assessment in a CI job

### DIFF
--- a/src/backend/setup.cfg
+++ b/src/backend/setup.cfg
@@ -1,0 +1,21 @@
+;;
+;; Third-party packages configuration
+;;
+[flake8]
+max-line-length = 99
+exclude =
+    .git,
+    .venv,
+    build,
+    venv,
+    __pycache__,
+    node_modules,
+    */migrations/*
+
+[isort]
+known_django=django
+include_trailing_comma=True
+line_length=88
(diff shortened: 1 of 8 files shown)


src/backend/setup.cfg
```diff
@@ -0,0 +1,22 @@
...
+    node_modules,
+    */migrations/*
+
+[isort]
+known_django=django
+known_richie=richie
+include_trailing_comma=True
+line_length=88
+multi_line_output=3
+sections=FUTURE,STDLIB,DJANGO,THIRDPARTY,RICHIE,FIRSTPARTY,LOCALFOLDER
```
sampaccoud: Here `richie` is a third party and `funmooc` should be isolated?

jmaupetit: I had hesitations on this point. You are probably right. I'll fix this.

jmaupetit: Now we consider that Richie must be a third party app, and we will use relative import for `funmooc` (_i.e._ project-related) imports.

PR #26, opened 2019-08-20
Implement SSO using an OpenEdx oauth2 backend

jmaupetit:
## Purpose

Fun-mooc user accounts should be shared among our services (Richie, LMS, CMS, ...), we thus need an SSO implementation using the LMS as an oauth2 provider (this is where users are registered for now).

## Proposal

- [x] add an hawthorn LMS instance to the development stack (oauth2 provider)
- [x] add an oauth2 authentication backend to fun-mooc
- [x] add an OpenID authentication backend to support older OpenEdx releases (at least dogwood)


Dockerfile
```diff
@@ -80,23 +80,12 @@ COPY requirements/dev.txt /tmp/requirements.txt
 # Install development dependencies
 RUN pip install -r /tmp/requirements.txt
 
-# Install dockerize. It is used to ensure that the database service is accepting
```
sampaccoud: In your commit message:
 
> It appears more effective

Can you explicit why in the commit message?

src/backend/funmooc/auth/backends.py
```diff
@@ -0,0 +1,263 @@
+"""Django authentication backends.
+
+For more information visit
+https://docs.djangoproject.com/en/dev/topics/auth/customizing/.
+
+This authentication backend is inspired from the official edx's auth-backends
```
sampaccoud:
```suggestion
This authentication backend is inspired by the official edX auth-backends
```

src/backend/funmooc/auth/templates/registration/login.html
```diff
@@ -0,0 +1,14 @@
...
+  <form method="post">
+    {% csrf_token %}
+    {{ form.as_p }}
+    <button type="submit">Login</button>
+  </form>
+  <br>
+  <p><strong>-- OR --</strong></p>
+  <a href="{% url 'social:begin' 'edx-oauth2' %}">Login with edX (oauth2)</a><br>
+  <a href="{% url 'social:begin' 'edx-oidc' %}">Login with edX (oidc)</a><br>
+{% endblock %}
```
sampaccoud: add i18n to this view?

But anyway we must bypass it for our login workflow right? I think I saw a setting for this in OpenedX

src/backend/funmooc/settings.py
```diff
@@ -206,6 +231,9 @@ class Base(DRFMixin, RichieCoursesConfigurationMixin, Configuration):
         "parler",
         "rest_framework",
         "storages",
+        "social_django",
+        # Fun mooc
+        "funmooc.auth.apps.AuthConfig",
```
sampaccoud: What is the point of declaring this as an app?

Dockerfile
jmaupetit: I think I'm explaining why in the commit message. Not sure to understand what you mean.

src/backend/funmooc/auth/templates/registration/login.html
jmaupetit: AFAIU this login view cannot be bypassed, the only thing we can bypass _via_ OpenEdx LMS admin interface is the authorization agreement view for a registered app.

For now this template is a POC and it deserves more love (i18n, styles, etc.), but more important than that, we should agree on the login logic depending on the targetted LMS release, _i.e._ oauth2 for hawthorn and oidc for dogwood (+eucalyptus). We should discuss about this.

src/backend/funmooc/settings.py
jmaupetit: I also think it's useless. I've left this from the first implementation.

jmaupetit: After thought: this app should be declared so that the template loader can find the `login.html` registration template.

sampaccoud:
We decided to include the social authentication backends in richie.core as they will be used by most projects. See [corresponding pull request](https://github.com/openfun/richie/pull/1057).

PR #41, opened 2019-10-04
Fix search page behind htaccess

sampaccoud:
## Purpose

The DRF API calls were responding with a 403 status code for anonymous users, and were working for authenticated users. This was due to a Basic Auth fallback option in the default configuration of DRF's authentication backends.

For an anonymous user:
- The first authentication backend in the list of DRF's default backends (Django session authentication) does not try to authenticate because it does not see any cookie,
- The second backend (Basic Authentication) tries to authenticate because it sees an Authentication header. But this header is coming from the Nginx htaccess (and is not known by Django), so the authentication fails with a 403.

Removing the Basic Authentication backend allows anonymous users to be really considered as anonymous in Django.

## Proposal

- Remove Basic Auth fallback option from DRF's authentication backends,
- Add pipework to run tests via make and in the CI,
- Add a test to validate that the site now works behind an htaccess.

Another alternative would have been that nginx does not pass the Basic Authentication headers. This can be done in [arnold](https://github.com/openfun/arnold) but we also don't want to allow basic authentication on our site so it had to be removed here as well.

### DIFF
--- a/src/backend/funmooc/base/tests/test_settings.py
+++ b/src/backend/funmooc/base/tests/test_settings.py
@@ -1,0 +1,50 @@
+"""Test fun-mooc configuration."""
+from unittest import mock
+
+from django.conf import settings
+from django.test import TestCase, override_settings
+
+from rest_framework.response import Response
+from rest_framework.settings import DEFAULTS, api_settings
+from richie.apps.search.viewsets.courses import CoursesViewSet
+
+
+class ConfigurationTestCase(TestCase):
+    """Validate that our configuration works as expected."""
+
+    @mock.patch.object(CoursesViewSet, "list", spec=True, return_value=Response({}))
+    def test_configuration_restframework_htaccess(self, _mock_list):
+        """The search API endpoint should work behind an htaccess."""
+        # First, check that API calls were broken with the default DRF configuration
+        # What was happening is that DRF defines Basic Authentication as a fallback by default
+        # and our query has a basic auth header with the username and password of the htaccess
+        # defined in nginx. Django was trying to authenticate a user with these credentials,
+        # which of course failed.
+        with override_settings(
+            REST_FRAMEWORK={
+                **settings.REST_FRAMEWORK,
+                "DEFAULT_AUTHENTICATION_CLASSES": DEFAULTS[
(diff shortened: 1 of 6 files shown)


src/backend/funmooc/base/tests/test_settings.py
```diff
@@ -0,0 +1,22 @@
...
+    def test_configuration_restframework_htaccess(self):
+        """The search API endpoint should work behind an htaccess."""
+        with mock.patch.object(
+            CoursesViewSet, "list", spec=True, return_value=Response({})
+        ):
+            response = self.client.get(
+                "/api/v1.0/courses/",
+                HTTP_AUTHORIZATION="Basic dXNlcm5hbWU6cGFzc3dvcmQ=",
+            )
+        self.assertEqual(response.status_code, 200)
```
jmaupetit: I have the feeling that this test is quite weak: 

1. IMO we need to test at least two conditions, _i.e._ with and without basic auth? 
2. Mocking the response you are testing and be happy that the mock returns a 200 response is... surprising! Am i missing something?
3. Where does the HTTP Basic auth token comes from?

sampaccoud: I don't think the test was weak but the explanation was. I added a counter test as you suggested and tried to explain the problem better in the test and in the PR description.

sampaccoud:
@jmaupetit please the PR description again and you should now understand. Sorry about that.

src/backend/funmooc/base/tests/test_settings.py
```diff
@@ -0,0 +1,50 @@
...
+from richie.apps.search.viewsets.courses import CoursesViewSet
+
+
+class ConfigurationTestCase(TestCase):
+    """Validate that our configuration works as expected."""
+
+    @mock.patch.object(CoursesViewSet, "list", spec=True, return_value=Response({}))
+    def test_configuration_restframework_htaccess(self, _mock_list):
+        """The search API endpoint should work behind an htaccess."""
+        # First check that API calls were broken with the default DRF configuration
```
jmaupetit:
```suggestion
        # First, check that API calls were broken with the default DRF configuration.
```

PR #50, opened 2020-04-20
⬆️(project) upgrade to richie 2.0.0-beta.3

sveetch:
We want to upgrade to the last richie version which contain the new
integration so we can finally add the fun-corporate theme.

This update the Python requirements, the frontend requirements,
settings and base urls.


src/backend/funmooc/base/admin.py
```diff
@@ -9,7 +10,7 @@
 from filer.models.virtualitems import FolderRoot
 
 
-class CustomUserAdmin(auth.admin.UserAdmin):
+class CustomUserAdmin(UserAdmin):
```
sampaccoud, line 13: what's the point of this?

src/backend/funmooc/urls.py
```diff
@@ -32,7 +33,15 @@
 urlpatterns += i18n_patterns(
     url(r"^admin/", admin.site.urls),
     url(r"^accounts/", include("django.contrib.auth.urls")),
-    url(r"^", include("cms.urls"))  # NOQA
+    url(
```
sampaccoud, line 36: fun-mooc is a real site not like richie. So we don't want the styleguide if DEBUG is False

src/backend/funmooc/base/admin.py
sveetch: It's a fix, possibly because a change in last Django version. When i launched the demo-site command it failed on this code arguing there was no UserAdmin in auth.admin.

src/backend/funmooc/settings.py
```diff
@@ -17,25 +17,77 @@
...
     except FileNotFoundError:
-        pass
+        return "NA"  # Default: not available
+
+
+class StyleguideMixin:
+    """
+    Theme styleguide reference
+
+    Only used to build styleguide page without to hardcode properties and
```
sampaccoud, line 41:
```suggestion
    Only used to build styleguide page without the need to hardcode properties and
```

PR #66, opened 2020-06-02
♻️(project) reorganize project to make it a site factory

sampaccoud:
## Purpose

Inspired by `openfun/openedx-docker` and `openfun/openedx-theme-factory`, we want to  turn this repository into a site factory to develop, lint and test all our sites based on Richie, as well as automate the building and publishing of their Docker images.

## Proposal

- [x] move funmooc's site sources to a `sites/funmooc` directory adapting docker-compose and Makefile to make it work
- [x] move aws code to project root and factorize between sites
- [x] refactor circleci configuration using the CLI pack tool and a custom bin/ci script to generate a config.yml file that automatically adapts to the projet's sites and their state


.circleci/config.yml
```diff
@@ -1,413 +1,367 @@
-generate-version-file: &generate-version-file
```
sampaccoud, line 1: Don't review this file as it is auto-generated.

.circleci/src/@parameters.yml
```diff
@@ -0,0 +1,5 @@
+# /!\ DO NOT EDIT: this file is autogenerated
```
sampaccoud, line 1: Don't review this file as it is autogenerated

.circleci/src/workflows/@funmooc.yml
```diff
@@ -0,0 +1,75 @@
+# /!\ DO NOT EDIT: this file is autogenerated
```
sampaccoud, line 1: Don't review this file as it is autogenerated

Makefile
```diff
@@ -44,6 +44,12 @@ build: ## build all containers
 	$(COMPOSE) build app-dev
 .PHONY: build
 
+reset:  ## Remove postgresql database
```
jmaupetit, line 47:
```suggestion
reset:  ## Remove database and local files
```

Makefile
```diff
@@ -44,6 +44,12 @@ build: ## build all containers
...
+	$(COMPOSE) stop
+	rm -Ir data/* || exit 0
```
jmaupetit, line 49: If you are trying to exit when expecting a `no` answer, this will not work as `rm` exit code will be `0` anyway.

aws/state.tf
```diff
@@ -0,0 +1,6 @@
+
```
jmaupetit, line 1: Is this first empty raw required?

bin/state
```diff
@@ -1,9 +1,11 @@
...
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
+
 # Run Terraform commands on the create_state_bucket project in the Hashicorp
 # docker container, passing our environment variables
 DOCKER_USER="$(id -u):$(id -g)" \
     docker-compose run --rm \
-        -v "${PWD}/src/aws/create_state_bucket:/app"\
-        terraform "$@"
+    -v "${PWD}/src/aws/create_state_bucket:/app"\
+    terraform "$@"
```
jmaupetit, line 11: I find previous indentation more readable.

docker-compose.yml
```diff
@@ -21,13 +21,17 @@ services:
       target: development
       args:
         DOCKER_USER: ${DOCKER_USER:-1000}
-    image: "funmooc:development"
-    env_file: "env.d/development"
+        SITE: ${ACTIVE_SITE:-funmooc}
+    image: "richie:development"
```
jmaupetit, line 25:
```suggestion
    image: "${ACTIVE_SITE:-funmooc}:development"
```

docker-compose.yml
```diff
@@ -39,11 +43,14 @@ services:
       target: production
       args:
         DOCKER_USER: ${DOCKER_USER:-1000}
+        SITE: ${ACTIVE_SITE:-funmooc}
     # We tag our images with the current commit sha1 in the CI to make them
     # unique and avoid collisions in parallel builds.
-    image: "funmooc:production"
-    env_file:
-      - env.d/production
+    image: "richie:production"
```
jmaupetit, line 49:
```suggestion
    image: "${ACTIVE_SITE:-funmooc}:production"
```

docker-compose.yml
```diff
@@ -61,7 +68,8 @@ services:
         DOCKER_USER: ${DOCKER_USER:-1000}
         NGINX_IMAGE_NAME: ${NGINX_IMAGE_NAME:-fundocker/openshift-nginx}
         NGINX_IMAGE_TAG: ${NGINX_IMAGE_TAG:-1.13}
-    image: "funmooc-nginx:production"
+        SITE: ${ACTIVE_SITE:-funmooc}
+    image: "richie-nginx:production"
```
jmaupetit, line 72:
```suggestion
    image: "${ACTIVE_SITE:-funmooc}-nginx:production"
```

bin/activate
```diff
@@ -0,0 +1,47 @@
...
+if [[ ${choice} -gt n_sites ]]; then
+    (>&2 echo "Invalid choice ${choice} (should be <= ${n_sites})")
+    exit 10
+fi
+
+if [[ ${choice} -le 0 ]]; then
+    choice=${default}
+fi
+
+ACTIVE_SITE="${sites[$((choice-1))]}"
```
jmaupetit, line 40: Why not using the same semantic as in the `Dockerfile`, _e.g._ `SITE` instead of `ACTIVE_SITE`?

sites/funmooc/VERSION
```diff
@@ -0,0 +1 @@
+0.10.1
```
jmaupetit, line 1: Why not using the following files instead:

- `sites/${SITE}/src/backend/setup.cfg` and/or `sites/${SITE}/src/backend/__init__.py`
- `sites/${SITE}/src/frontend/package.json`

aws/state.tf
sampaccoud: no I removed it :+1:

bin/state
sampaccoud: completely... this was not intentional.

docker-compose.yml
sampaccoud: well I don't really care how this image is called and I prefer to not pollute my system with too many different image names. I'm ok to do that if you think it's important.

bin/activate
sampaccoud: Because this is exported so I thought the name should be different and more explicit.

docker-compose.yml
sampaccoud: And it will also complicate the CI configuration as I will need to variabilize all image names.

Makefile
```diff
@@ -62,6 +71,11 @@ stop: ## stop the development server
 	@$(COMPOSE) stop
 .PHONY: stop
 
+info:  ## get activated site info
+	@echo "ACTIVE_SITE: $(COLOR_INFO)$(ACTIVE_SITE)$(COLOR_RESET)"
```
jmaupetit, line 75: `COLOR_RESET` is not defined.

bin/ci
```diff
@@ -0,0 +1,109 @@
...
+# Usage: update_sources
+function update_sources() {
+    # Reset the content of the .circleci/src/@root.yml file
+    disclaimer="# /!\ DO NOT EDIT: this file is autogenerated"
+    printf "%s\nparameters:\n" "${disclaimer}" > ".circleci/src/@parameters.yml"
+
+    while IFS= read -r -d '' site_path
+    do
+        # shellcheck disable=SC2001
+        site=$(echo "${site_path}" | sed 's|'"${SITES_DIRECTORY}\/"'||')
```
jmaupetit, line 38: Use parameter substitution instead, `shellcheck` will be happy with this:

```suggestion
        site=$(echo "${site_path//${SITES_DIRECTORY}\//}")
```

bin/ci
```diff
@@ -0,0 +1,109 @@
...
+
+    # Check config validity
+    docker run --rm -v "${PROJECT_DIRECTORY}:/data" circleci/circleci-cli:alpine \
+        config validate /data/.circleci/config.yml
+}
+
+
+# Utility function that checks if the circle configuration committed to git
+# corresponds to the state of the project so that the workflows and jobs are
+# run according to what has changed as compred to the "master" branch.
```
jmaupetit, line 80:
```suggestion
# run according to what has changed as compared to the "master" branch.
```

bin/ci
```diff
@@ -0,0 +1,109 @@
...
+#
+# Usage: check_configuration
+function check_configuration() {
+    update_sources
+    if \
+        git ls-tree master --name-only | grep .circleci/config.yml && \
+        git diff --name-only | grep "^.circleci/config.yml$";
```
jmaupetit, line 87: You should use `update` first then and not `update_sources` or else, you may have changed sources but forgot to pack them and this won't be detected (as the `.circleci/config.yml` file may not have been changed).

jmaupetit, line 87: Alternative implementation:

```suggestion
        git ls-tree master --name-only | grep "^.circleci/src" && \
        ! git ls-tree master --name-only | grep "^.circleci/config.yml$";
```

docker-compose.yml
jmaupetit: Yes, this is one of my review command. It' not that complicated IMO.

This is a label, it won't change cache layers. I think it's more relevant to have a proper name for your images as `richie` could target the core project development or any derived site...

bin/terraform
```diff
@@ -1,7 +1,22 @@
...
+        "-backend-config=dynamodb_table=${ACTIVE_SITE}_terraform_state_locks"
+    )
+fi
+
 # Run Terraform commands in the Hashicorp docker container, passing our
 # environment variables
-DOCKER_USER="$(id -u):$(id -g)" \
-    docker-compose run --rm terraform "$@"
+DOCKER_USER="$(id -u):$(id -g)" docker-compose run --rm terraform \
+    "${args[@]}"
```
jmaupetit, line 22: I find previous indentation more readable.

docs/aws.md
```diff
@@ -31,17 +32,41 @@ TF_VAR_aws_region=eu-west-1
 > i. S3 buckets, ii. KMS keys, iii. DynamoDB tables, iv. IAM users and v.
 > CloudFront distributions.
 
+## Project settings
+
+You need to configure domain names for the CDN in a setting file specific to
+your project: `sites/{yourSite}/aws/configtfvars`:
```
jmaupetit, line 38:
```suggestion
your project: `sites/{{ your_site }}/aws/configtfvars`:
```

README.md
```diff
@@ -1,28 +1,46 @@
...
 ```
 
-The easiest way to start working on the project is to use our `Makefile`:
+## Getting started
+
+First, you need to activate the site on which you want to work. We provide
+a script that will list existing sites and let you choose one. Source it:
 
+```bash
+$ . bin/activate
```
jmaupetit, line 26: This won't work if user's default shell is not BASH. I would recommend to execute it instead:

```suggestion
$ bin/activate
```

README.md
```diff
@@ -1,28 +1,46 @@
...
+Select an available site to activate:
+[1] demo (default)
+[2] funmooc
+Your choice: 2
+
+# Check your environment with:
+make info
```
jmaupetit, line 33:
```suggestion
$ make info
```

Makefile
```diff
@@ -62,6 +71,11 @@ stop: ## stop the development server
...
+	@echo ""
```
jmaupetit, line 76: Is this empty row required?

bin/activate
jmaupetit: `RICHIE_SITE` would be more explicit then?

.circleci/src/jobs/@backend.yml
```diff
@@ -0,0 +1,98 @@
+# Backend jobs
+# Build, lint and test backend development environment for a site
+build-back:
+  parameters:
+    site:
+      type: string
+  docker:
+    - image: circleci/python:3.7-stretch
```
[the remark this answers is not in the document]
sampaccoud: I kept it the way it was done but if we want to reuse, I think we should use "executors". Should be done in other PR though.

.circleci/src/jobs/@docker.yml
```diff
@@ -0,0 +1,108 @@
...
+    # Activate docker-in-docker (with layers caching enabled)
+    - setup_remote_docker:
+        docker_layer_caching: true
+    - run:
+        name: Build docker images
+        command: make env.d/aws && make build
+    - run:
+        name: Check built image availability
+        command: |
+            docker images "richie:development"
```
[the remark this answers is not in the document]
sampaccoud: Yeah I think you're right... let's do that.

sites/funmooc/VERSION
sampaccoud: good idea

Makefile
sampaccoud: no

Dockerfile
```diff
@@ -114,7 +123,7 @@ ARG DOCKER_USER
 USER ${DOCKER_USER}
 
 # The default command runs gunicorn WSGI server in the sandbox
-CMD gunicorn -c /usr/local/etc/gunicorn/funmooc.py funmooc.wsgi:application
+CMD gunicorn -c /usr/local/etc/gunicorn/app.py app.wsgi:application
```
jmaupetit, line 126:
```suggestion
ARG SITE
CMD gunicorn -c /usr/local/etc/gunicorn/app.py ${SITE}.wsgi:application
```
Or rename the `sites/${SITE}/src/backend/${SITE}` folder to `app` and `DJANGO_SETTINGS_MODULE` should also follow.

sampaccoud: yep so we have to enforce the name of this app :man_shrugging: so better force it to something constant like `app`.

sampaccoud: no.. better call it {SITE} as per Django common practice.

bin/activate
sampaccoud: ok

bin/ci
sampaccoud: It works!

sites/funmooc/VERSION
sampaccoud: ok... so since I'm already checking for modification to sites/${SITE}/src/backend we are already tracking changes on the backend changes.

I will add a check in the CI that the tag, backend and the frontend version are the same.

bin/ci
sampaccoud: true that

sampaccoud: In fact the first line is only here to avoid failing on the present PR where `.circleci/config.yml` does not exist yet.

I think we can fun `bin/ci update` and check that nothing has changed in the `.circleci` folder...

sampaccoud: @jmaupetit I think I nailed it but take a look :pray:

bin/terraform
sampaccoud: Well... shellcheck didn't like it so I found out that everything fits in one line...

Dockerfile
```diff
@@ -113,17 +113,18 @@ CMD python manage.py runserver 0.0.0.0:8000
...
 # Copy collected symlinks to static files
 COPY --from=collector ${STATIC_ROOT}/staticfiles.json ${STATIC_ROOT}/
 
 # Un-privileged user running the application
-ARG DOCKER_USER
 USER ${DOCKER_USER}
 
 # The default command runs gunicorn WSGI server in the sandbox
-CMD gunicorn -c /usr/local/etc/gunicorn/app.py app.wsgi:application
+CMD gunicorn -c /usr/local/etc/gunicorn/app.py ${SITE}.wsgi:application
```
jmaupetit, line 127: For the `CMD` statement, the `${SITE}` variable will be considered as an environment variable and will be interpolated by the shell, so you have to define an environment variable from the corresponding `ARG`:

```suggestion
ENV SITE=${SITE}
CMD gunicorn -c /usr/local/etc/gunicorn/app.py ${SITE}.wsgi:application
```

bin/ci
```diff
@@ -77,28 +76,42 @@ function update() {
...
 function check_configuration() {
-    update_sources
+    update
     if \
         git ls-tree master --name-only | grep .circleci/config.yml && \
-        git diff --name-only | grep "^.circleci/config.yml$";
+        git diff --name-only | grep "^.circleci/";
     then
-        echo "check: the .circleci/config.yml file does not reflect the latest changes."
+        echo "check: the .circleci configuration does not reflect the latest changes."
```
jmaupetit, line 88:
```suggestion
        echo "check: the .circleci configuration does not reflect latest changes."
```

bin/ci
```diff
@@ -77,28 +76,42 @@ function update() {
...
+# run according to what has changed as compared to the "master" branch.
 #
 # Usage: check_configuration
 function check_configuration() {
-    update_sources
+    update
     if \
         git ls-tree master --name-only | grep .circleci/config.yml && \
-        git diff --name-only | grep "^.circleci/config.yml$";
+        git diff --name-only | grep "^.circleci/";
```
jmaupetit, line 86: Finally, you only need this test once you have run the `update` function? Then I would remove the first test, _i.e._

```git ls-tree master --name-only | grep .circleci/config.yml &&```

File: bin/ci [hunk 28]
```diff
@@ -77,28 +76,42 @@ function update() {
...
+# argument is properly set both in the backend and frontend code.
+#
+# Usage: check_tag SITE TAG
+function check_tag() {
+    site=$1
+    src="${SITES_DIRECTORY}/${site}/src"
+    tag=$2
+
+    grep "__version__ = \"${tag}\"" "${src}/backend/${site}/__init__.py" &> /dev/null
+    grep "\"version\": \"${tag}\"" "${src}/frontend/package.json" &> /dev/null
```
jmaupetit, line 104: Having an explicit error message and exit code would be nice?

```suggestion
    declare -i fail
    fail=0
    if ! grep "__version__ = \"${tag}\"" "${src}/backend/${site}/__init__.py" &> /dev/null ; then
      echo "Backend version does not match the current tag!"
      fail+=1
    fi
    if ! grep "\"version\": \"${tag}\"" "${src}/frontend/package.json" &> /dev/null ; then
      echo "Frontend version does not match the current tag!"
      fail+=2
    fi
    exit ${fail}
```

bin/ci
sampaccoud: True!

PR #68, opened 2020-06-04
Add cnfpt site and refactor circleci config to work well with Github status checks

sampaccoud:
## Purpose

We want to add a new site for https://mooc.cnfpt.fr, porting it from https://github.com/openfun/richie-cnfpt.

## Proposal

I copy/pasted the project from github.com/openfun/richie-cnfpt and adapted it to new custom nginx image and way to build with the site factory.

### Refactor CircleCI workflows

We were only running workflows for projects that required it but in github, status checks can only be marked as durably required or not. Workflows had to be marked required when they were run and not required when they  were absent... This is not possible so pull requests would be blocked.

I propose to always run all workflows, mark them as required in github, and run a dummy job if nothing has changed.


.circleci/src/jobs/@frontend.yml
```diff
@@ -12,8 +12,8 @@ build-front-production:
         path: ~/fun
     - restore_cache:
         keys:
-          - v2-front-dependencies-{{ checksum "yarn.lock" }}
-          - v2-front-dependencies-
+          - v3-front-dependencies-{{ checksum "yarn.lock" }}
+          - v3-front-dependencies-
```
jmaupetit, line 16: :thinking: maybe the site name should be included in cache keys?

sampaccoud: Done

sampaccoud:
@jmaupetit you need to review the latest commits as I refactored workflows to work well with Github status checks (see commit message or PR description for explanations)

PR #69, opened 2020-06-05
Add richie demo site and allow overriding Docker image name

sampaccoud:
### Purpose

We want to add a "richie-demo" Docker image to replace the image that is built in richie's repository at github.com/openfun/richie.

### Proposal

- [x] Add a "demo" site with minimal customizations,
- [x] Allow overriding the Docker image name.
- [x] Remove sites' workflow files from the git history as they are only intermediary files that don't hold any new information and are polluting reviews.


bin/ci
```diff
@@ -51,8 +51,12 @@ function update_sources() {
             template=".circleci/src/workflows/site_no_change.yml.tpl"
         fi
 
+        # Look for custom image name in site's setup.cfg file and default to the site name
+        image=$(grep -Po '(?<=^name = ).*' "${site_path}/src/backend/setup.cfg" || echo "${site}")
```
jmaupetit, line 55: Wouldn't it be easier to read with a grep/sed combo instead of a PCRE?

jmaupetit: Or even grep/cut:

```suggestion
        image=$(grep "^name = " "${site_path}/src/backend/setup.cfg" | cut -c 7- || echo "${site}")
```

sampaccoud: you're right it's more readable!

.circleci/src/workflows/site-factory.yml
```diff
@@ -0,0 +1,18 @@
+jobs:
```
sampaccoud, line 1: @jmaupetit I'm moving this file inside a folder to allow git ignoring the workflow files related to sites.

bin/ci
```diff
@@ -95,18 +100,30 @@ function check_configuration() {
...
+    init_path="${site_path}/src/backend/${site}/__init__.py"
+    if ! grep "__version__ = \"${version}\"" "${init_path}" &> /dev/null ; then
       echo "Backend version does not match the current tag!"
       fail+=1
     fi
-    if ! grep "\"version\": \"${tag}\"" "${src}/frontend/package.json" &> /dev/null ; then
+    if ! grep "\"version\": \"${version}\"" "${site_path}/src/frontend/package.json" &> /dev/null ; then
       echo "Frontend version does not match the current tag!"
-      fail+=2
+      fail+=1
```
jmaupetit, line 126: At first, `fail` was not a counter but a way to have a unique exit code for an error combination. Now that you have many error, maybe it should be a fixed code instead, _e.g._

```suggestion
      fail=4
```

sampaccoud: done

PR #76, opened 2020-06-12
Finalize terraform project to deploy to aws

sampaccoud:
### Purpose

Trying to deploy all sites to AWS, I had to make some adjustments.

### Proposal

- [x] Upgrade to Terraform 0.12.26
- [x] Share the same bucket to store the state of all sites
- [x] Set custom location to each site's folder for the `terraform.tfstate` file

### DIFF
--- a/bin/state
+++ b/bin/state
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-# shellcheck source=bin/_config.sh
-source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
-
 # Run Terraform commands on the create_state_bucket project in the Hashicorp
 # docker container, passing our environment variables
+
+
+docker-compose pull terraform-state
+
+# Run Terraform commands in the Hashicorp docker container via docker compose
 DOCKER_USER="$(id -u):$(id -g)" \
-    docker-compose run --rm \
-        -v "${PWD}/src/aws/create_state_bucket:/app"\
-        terraform "$@"
+    docker-compose run --rm terraform-state "$@"
--- a/bin/terraform
+++ b/bin/terraform
@@ -6,16 +6,18 @@ source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
 
 args=( "$@" )
 
-args+=("-var-file=/config/config.tfvars")
+case "$1" in
+    apply|destroy|import|plan)
+        args[0]="${args[0]} -var-file=/config/config.tfvars"
+        ;;
+    init)
+        args[0]="${args[0]} -backend-config=key=${RICHIE_SITE}.tfstate"
+        ;;
+esac
 
-if [[ "$1" == "init" ]] ; then
-    args+=(
-        "-backend-config=bucket=${RICHIE_SITE}-terraform"
-        "-backend-config=key=${RICHIE_SITE}.tfstate"
-        "-backend-config=dynamodb_table=${RICHIE_SITE}_terraform_state_locks"
-    )
-fi
+docker-compose pull terraform
 
-# Run Terraform commands in the Hashicorp docker container, passing our
-# environment variables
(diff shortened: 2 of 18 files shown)


bin/terraform
```diff
@@ -6,16 +6,17 @@ source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
...
+        args[0]="${args[0]} -backend-config=key=${RICHIE_SITE}.tfstate"
+        ;;
+esac
 
 # Run Terraform commands in the Hashicorp docker container, passing our
 # environment variables
-DOCKER_USER="$(id -u):$(id -g)" docker-compose run --rm terraform "${args[@]}"
+# shellcheck disable=SC2068
+DOCKER_USER="$(id -u):$(id -g)" \
+    docker-compose run --rm terraform ${args[@]}
```
sampaccoud, line 22: @jmaupetit shellcheck is not happy with this but it doesn't work if I do:
```suggestion
    docker-compose run --rm terraform "${args[@]}"
```
So I ignored it but if you have a better solution don't hesitate!

bin/state
```diff
@@ -1,12 +1,13 @@
...
-        -v "${PWD}/src/aws/create_state_bucket:/app"\
-        terraform "$@"
+
+docker run --rm -it \
+    --env-file ./env.d/aws \
+    -u "$(id -u)" \
+    -v "${PWD}/aws/create_state_bucket:/app" \
+    -w "/app" \
+    "hashicorp/terraform" \
+    "$@"
```
jmaupetit, line 13: Why not using the docker-compose `terraform` service for this?

bin/terraform
jmaupetit: I think you are right on this: you explicitly want to split array elements or else docker-compose cannot correctly interpret your command.

docker-compose.yml
```diff
@@ -89,9 +89,10 @@ services:
       - ./sites/${RICHIE_SITE:-funmooc}:/app
 
   terraform:
-    image: hashicorp/terraform:light
+    image: hashicorp/terraform
```
jmaupetit, line 92: Why not pinning a release instead?

bin/state
sampaccoud: I don't know I did not find an advantage to use docker-compose. The thing is I can't use the existing `terraform` service in docker-compose because if defines environment variables that I don't want here.

bin/terraform
sampaccoud: :+1:

docker-compose.yml
sampaccoud: Because I like to be forced to upgrade. Terraform upgrades are usually very easy and we would never upgrade if we pinned it :man_shrugging:

jmaupetit: :thinking: ok then.

bin/state
jmaupetit: Those variables are potentially harmful or without any effect for what you are trying to achieve?

sampaccoud: yes they should not be set when running bin/state

docker-compose.yml
sampaccoud: well... we will always use the latest version available... no going back. I'm ok to pin if you think it's problematic but I'm afraid we will never upgrade.

lunika: We should find a way to docker pull the latest version available ? Maybe doint it in one of the `bin/*` script ? Be cause after a first bootstrap we will never update it.

jmaupetit: Maybe we should force latest image pulling by adding `docker-compose pull terraform` some where in our `Makefile` or `bin/*` scripts?

The best solution would have been a pull policy in the `docker-compose` file, but it has not been implemented yet.

sampaccoud: @lunika @jmaupetit I added a fixup to use docker-compose also for the terraform state script and made sure that the latest images are always up-to-date before running terraform.

PR #82, opened 2020-07-15
🐛(Makefile) fix database directory initialization

madmatah:
## Purpose

I want to bootstrap richie by following the **Getting started** instructions from the README.

The `make bootstrap` command failed with the following error (during the `make run` step) : 

```
Starting richie-site-factory_app-dev_1 ... done
2020/07/15 11:21:44 Waiting for: tcp://db:5432
2020/07/15 11:21:44 Problem with dial: dial tcp: lookup db on 127.0.0.11:53: no such host. Sleeping 1s
2020/07/15 11:21:45 Problem with dial: dial tcp: lookup db on 127.0.0.11:53: no such host. Sleeping 1s
[...]
2020/07/15 11:22:43 Problem with dial: dial tcp: lookup db on 127.0.0.11:53: no such host. Sleeping 1s
2020/07/15 11:22:44 Timeout after 1m0s waiting on dependencies to become available: [tcp://db:5432]
make: *** [Makefile:80: run] Error 1
```

After looking at log files, I found that the database failed to start because of the following error : 

```
docker-compose logs db

db_1               | 
db_1               | initdb: directory "/var/lib/postgresql/data" exists but is not empty
db_1               | It contains a dot-prefixed/invisible file, perhaps due to it being a mount point.
db_1               | Using a mount point directly as the data directory is not recommended.
db_1               | Create a subdirectory under the mount
```

The database volume directory is initialized with a `.keep` file and the postgresql container fails to start because of this file.

## Proposal

Initialize the database volume with an empty directory.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,359 +1,358 @@
 commands:
-  generate-version-file:
-    parameters:
-      site:
-        default: funmooc
-        type: string
-    steps:
-    - run:
-        command: |
-          # Create a version.json à-la-mozilla
-          # https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md
-          printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' \
-          "$CIRCLE_SHA1" \
-          "$CIRCLE_TAG" \
-          "$CIRCLE_PROJECT_USERNAME" \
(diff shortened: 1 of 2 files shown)


sampaccoud:
@madmatah Still wondering why I did not get this error on my side :man_shrugging: . Concerning the CI error you get, you must run `bin/ci update` to synchronize the CircleCI configuration with the project.

PR #90, opened 2020-08-12
Enable cache with redis-sentinel for all sites

madmatah:
## Purpose

We want to improve performances of our sites by enabling content and session cache.

## Proposal

I made a POC with the `cnfpt` site.
When approved, we should do this on all sites.

- [x] Upgrade richie to a version > 2.0.0-beta.9 (with sentinel support)
- [x] Monkey patch Django CMS to enable page cache for non-staff users
- [x] Enable Django cache (`CACHES` setting + middlewares)
- [x] Increase Django CMS cache duration
- [x] Use cache for sessions (with the [`cached`](https://docs.djangoproject.com/en/3.0/topics/http/sessions/#using-cached-sessions) `SESSION_ENGINE`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ COMPOSE_TEST_RUN     = $(COMPOSE) run --rm -e DJANGO_CONFIGURATION=Test
 COMPOSE_TEST_RUN_APP = $(COMPOSE_TEST_RUN) app-dev
 WAIT_DB              = $(COMPOSE_RUN) dockerize -wait tcp://db:5432 -timeout 60s
 WAIT_ES              = $(COMPOSE_RUN) dockerize -wait tcp://elasticsearch:9200 -timeout 60s
+WAIT_SENTINEL        = $(COMPOSE_RUN) dockerize -wait tcp://redis-sentinel:26379 -wait tcp://redis-primary:6379 -timeout 20s
 
 # -- Node
 
@@ -75,6 +76,8 @@ logs: ## display app logs (follow mode)
 .PHONY: logs
 
 run: ## start the wsgi (production) or development server
+	@$(COMPOSE) up -V -d redis-sentinel
+	@$(WAIT_SENTINEL)
 	@$(COMPOSE) up -d nginx
 	@$(COMPOSE) up -d app-dev
 	@$(WAIT_DB)
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,8 @@ lint-back: \
   lint-back-black \
   lint-back-flake8 \
   lint-back-pylint \
-  lint-back-bandit
+  lint-back-bandit \
+  lint-back-raincoat
 .PHONY: lint-back
 
 lint-back-black: ## lint back-end python sources with black
@@ -178,6 +179,11 @@ lint-back-pylint: ## lint back-end python sources with pylint
 	@echo 'lint:pylint started…'
 	@$(COMPOSE_TEST_RUN_APP) pylint .
 .PHONY: lint-back-pylint
+
+lint-back-raincoat: ## lint back-end python sources with raincoat
+	@echo 'lint:raincoat started…'
+	@$(COMPOSE_TEST_RUN_APP) raincoat
+.PHONY: lint-back-raincoat
 
 lint-back-bandit: ## lint back-end python sources with bandit
 	@echo 'lint:bandit started…'
--- a/sites/cnfpt/src/backend/base/admin.py
+++ b/sites/cnfpt/src/backend/base/admin.py
@@ -5,9 +5,12 @@ from django.http import JsonResponse
 from django.utils.translation import ugettext_lazy as _
 from django.views.decorators.csrf import csrf_exempt
 
+import cms.cache.page
 from filer.admin import clipboardadmin
 from filer.admin.clipboardadmin import ajax_upload as filer_ajax_upload
 from filer.models.virtualitems import FolderRoot
+
+from .cache import set_page_cache
 
 
 class CustomUserAdmin(UserAdmin):
@@ -39,3 +42,13 @@ FolderRoot.virtual_folders = lambda o: []
 user_model = auth.get_user_model()
 admin.site.unregister(user_model)
 admin.site.register(user_model, CustomUserAdmin)
+
+# Django CMS disables page cache for authenticated users
+# (See https://github.com/divio/django-cms/blob/3.7.4/cms/cache/page.py#L38)
+#
+# In our case, since most of our users will be authenticated and be served the
+# same content, this is a problem. That's why we are monkey patching the
+# function `cms.cache.page.set_page_cache`.
+# All non-staff users will benefit from page cache.
+
+cms.cache.page.set_page_cache = set_page_cache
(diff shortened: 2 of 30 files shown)


Makefile
```diff
@@ -75,6 +77,10 @@ logs: ## display app logs (follow mode)
 .PHONY: logs
 
 run: ## start the wsgi (production) or development server
+	@$(COMPOSE) up -d redis-primary
+	@$(WAIT_REDIS)
+	@$(COMPOSE) up -d redis-sentinel
```
sampaccoud, line 82: Maybe it would make sense that the redis-sentinel service depend on the redis-primary service in docker-compose?

sites/cnfpt/src/backend/base/admin.py
```diff
@@ -39,3 +42,5 @@ def ajax_upload(request, folder_id=None):
 user_model = auth.get_user_model()
 admin.site.unregister(user_model)
 admin.site.register(user_model, CustomUserAdmin)
+
+cms.cache.page.set_page_cache = set_page_cache
```
sampaccoud, line 46: I think we need a comment to explain that we are monkey patching and why with a link to the problem in DjangoCMS code.

sites/cnfpt/src/backend/cnfpt/settings.py
```diff
@@ -370,6 +372,38 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
         default="richie", environ_name="RICHIE_ES_INDICES_PREFIX", environ_prefix=None
     )
 
+    # Cache
```
sampaccoud, line 375: All this can be removed as well when Richie is upgraded to a version > beta.9 right? Maybe it will be beneficial to release it before merging this one.

sites/cnfpt/src/backend/cnfpt/settings.py
```diff
@@ -370,6 +372,38 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
...
+    CACHES = {
+        "default": {
```
sampaccoud, line 377: Actually we need a separate cache for sessions so that we assign it to a different database in redis. I should probably have commented it on your PR in openfun/richie as well?

For projects with both `edxapp` and `richie`, we need to set different redis db numbers for edxapp and richie cache as well!

sites/cnfpt/src/backend/cnfpt/settings.py
madmatah: Yes, that's the plan !

Makefile
madmatah: The dependency is already defined in the docker-compose.
This was to make sure that `redis-primary` is reachable when `redis-sentinel` starts.

sites/cnfpt/src/backend/base/admin.py
madmatah: I added a comment to explain the monkey patch.

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,84 @@
+"""
+Plugin for django-redis that supports redis sentinel.
+Credits :
+ - https://github.com/lamoda/django-sentinel
+ - https://github.com/KabbageInc/django-redis-sentinel
```
sampaccoud, line 5: the docstring is no more relevant

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,84 @@
...
+from django.utils.timezone import now
+
+from cms.cache import _get_cache_version, _set_cache_version
+from cms.cache.page import _page_cache_key
+from cms.constants import EXPIRE_NOW, MAX_EXPIRATION_TTL
+from cms.toolbar.utils import get_toolbar_from_request
+from cms.utils.conf import get_cms_setting
+
+
+def set_page_cache(response):
```
sampaccoud, line 24: I'm thinking that this may be hard to maintain and maybe there is a way with a context manager to call the original function patching the user. Even more DYI but this is hopefully temporary waiting for DjangoCMS to accept our change. What do you think?

sites/cnfpt/src/backend/cnfpt/settings.py
madmatah: In our case, the only reason to separate richie's cache and sessions in different database would be to avoid key conflicts, but django cache key prefixes prevent that.

And we will be able to separate `edxapp` and `richie` cache entries in different databases with the LOCATION setting.

sampaccoud: IC. what about other parameters like default timeout?

sites/cnfpt/src/backend/base/cache.py
madmatah: I agree with you on this : this may be hard to maintain and we'll have to follow django CMS's changes carefully.

But I really don't like to patch the user before calling the original function. We'll patch an attribute that has nothing to do with caching and this might lead to dreadful debugging sessions if the behavior of the parent functions changes in the future or if we introduce a bug.

And I don't think it's worth submitting this patch to DjangoCMS. I understand their choice : in most cases, logged user are not served the exact same content. We are a special case. Maybe they could add an option to disable this, but it has to be handled **VERY** carefully to not expose any personal user data to other users because of the page cache.

sites/cnfpt/src/backend/cnfpt/settings.py
madmatah: You're right, `KEY_FUNCTION`, `KEY_PREFIX`, `TIMEOUT` and `VERSION` are missing here.
This should be added in `richie` also.

I wish I could have used the [CacheURLValue](https://django-configurations.readthedocs.io/en/latest/values/#configurations.values.CacheURLValue) of django-configurations (I tried !) but it does not support our use case.

What do you think about Wrapping the `CACHES` definition in a values.DictValue ? I don't know if it's possible : 

```python
CACHES = values.DictValue({
        "default": {
            "BACKEND": values.Value(
                "django_redis.cache.RedisCache",
                environ_name="CACHE_DEFAULT_BACKEND",
                environ_prefix=None,
            ),
            "LOCATION": values.Value(
                "mymaster/redis-sentinel:26379,redis-sentinel:26379/0",
                environ_name="CACHE_DEFAULT_LOCATION",
                environ_prefix=None,
            ),
            "OPTIONS": values.DictValue(
                {"CLIENT_CLASS": "richie.apps.core.cache.SentinelClient"},
                environ_name="CACHE_DEFAULT_OPTIONS",
                environ_prefix=None,
            ),
        }
    })
```

sampaccoud: I would say it should work :man_shrugging:

madmatah: Done, everything is overridable with environment variables.
And I added TIMEOUT because it's often used.

sites/cnfpt/src/backend/base/cache.py
madmatah: fixed

sites/cnfpt/src/backend/base/cache.py
madmatah: Following your suggestions, I added `raincoat` as a new linter :-)

sampaccoud (approved):

madmatah:
@sampaccoud I updated all sites, i'm ready for a final review :pray:

sampaccoud (approved):

PR #93, opened 2020-08-19
Fix cross domain requests for static files with a querystring

sampaccoud:
## Purpose

Some static files are called with a query string. It breaks the effect of CORS headers to allow cross domain requests when static files are placed behing a CDN.

## Proposal

Passing the query string to the origin fixes the issue.

### DIFF
--- a/aws/cloudfront.tf
+++ b/aws/cloudfront.tf
@@ -45,7 +45,7 @@ resource "aws_cloudfront_distribution" "richie_cloudfront_distribution" {
     target_origin_id = local.s3_static_origin_id
 
     forwarded_values {
-      query_string = false
+      query_string = true
       headers = ["Access-Control-Request-Headers", "Access-Control-Request-Method", "Authorization", "Origin"]
 
       cookies {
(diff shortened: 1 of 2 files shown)


aws/cloudfront.tf
```diff
@@ -45,7 +45,7 @@ resource "aws_cloudfront_distribution" "richie_cloudfront_distribution" {
     target_origin_id = local.s3_static_origin_id
 
     forwarded_values {
-      query_string = false
+      query_string = true
```
madmatah, line 48: According to [cloudfront documentation](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_ForwardedValues.html#cloudfront-Type-ForwardedValues-QueryString), this is deprecated and you should use an origin request policy instead.

madmatah: Apparently this is not yet merged into the aws terraform provider ( https://github.com/terraform-providers/terraform-provider-aws/issues/14373 ).

madmatah (approved):

aws/cloudfront.tf
sampaccoud: Good catch. Let's upgrade to 0.13 first.

PR #103, opened 2020-08-25
Customize social networks and filters for funcampus and funcorporate

sampaccoud:
### Purpose

We want to customize social networks links and search filters for funcampus and funcorporate.

### Proposal

For the moment, we use the same social network links as for funmooc.

On fun-campus, show only search filters on 4 dimensions of categories.
On fun-corporate, remove search filters related to course sessions that are not present on this site.

Note: I also added i18n compilation in the DockerFile so the translations are available out-of-the-box.

### DIFF
--- a/sites/funcampus/src/backend/templates/social-networks/blogpost-badges.html
+++ b/sites/funcampus/src/backend/templates/social-networks/blogpost-badges.html
@@ -1,0 +1,48 @@
+{% load i18n static %}{% spaceless %}
+{% with full_page_url=SITE.web_url|add:page_url %}
+    {% blocktrans with title=page_title asvar mailto_subject %}FUN Campus news: {{ title }}{% endblocktrans %}
+    {% blocktrans with title=page_title|safe url=full_page_url asvar shared_sentence %}FUN Campus news: {{ title }} {{ url }}{% endblocktrans %}
+    <div class="social-network-badges">
+        <a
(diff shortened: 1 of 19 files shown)


sites/funcampus/src/backend/templates/social-networks/blogpost-badges.html
```diff
@@ -0,0 +1,48 @@
+{% load i18n static %}{% spaceless %}
+{% with full_page_url=SITE.web_url|add:page_url %}
+    {% blocktrans with title=page_title asvar mailto_subject %}FUN Campus news: {{ title }}{% endblocktrans %}
+    {% blocktrans with title=page_title|safe url=full_page_url asvar shared_sentence %}FUN Campus news: {{ title }} {{ url }}{% endblocktrans %}
+    <div class="social-network-badges">
+        <a
+            href="https://www.facebook.com/share.php?u={{ full_page_url }}"
```
madmatah, line 7: No need to urlencode the `full_page_url` ?
(Multiple occurences of this in this PR)

sampaccoud: Apparently not and it is already in production on other websites.

madmatah: Ok then :-)

madmatah (approved):

PR #119, opened 2020-09-11
💄(funcorporate) fix course category badge layout

jbpenrath:
## Purpose
On funcorporate, there was a difference between mockups and integration on course detail page; category badges layout was not correct.

## Proposal

- [x]  Extend subheader styles to customize category badges displayed inside it.

Capture d’écran 2020-09-11 à 10 16 18

### DIFF
--- a/sites/funcorporate/CHANGELOG.md
+++ b/sites/funcorporate/CHANGELOG.md
@@ -17,7 +17,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Customize social networks links
 
 ### Fixed
-
+- Override .subheader .category-badge styles to fit fun-corporate mockups.
 - Add i18n messages compilation in the DockerFile so translations are ready
 
 ## [0.5.0] - 2020-08-20
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -327,66 +327,11 @@ version: 2.1
 workflows:
     cnfpt:
         jobs:
-            - check-changelog:
-                filters:
-                    branches:
-                        ignore: master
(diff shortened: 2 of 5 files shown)


sampaccoud:
@jbpenrath Do you think it is possible to do this with css overrides only? If necessary, we can add things on richie side to make it possible. We should avoid maintaining template overrides for this kind of theme customisation.

jbpenrath:
> @jbpenrath Do you think it is possible to do this with css overrides only? If necessary, we can add things on richie side to make it possible. We should avoid maintaining template overrides for this kind of theme customisation.

Ok, I will overrides css instead of overrides template.

jbpenrath:
@sampaccoud Job done! 🙂

sites/funcorporate/CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Customize social networks links
 
 ### Fixed
-
+- Update `course_detail` template to fit fun-corporate mockups.
```
sampaccoud, line 20:
```suggestion
- Override category badge style to fit fun-corporate mockups.
```

sampaccoud (approved):

sites/funcorporate/CHANGELOG.md
jbpenrath: J'y ai pas pushé 😅

jbpenrath: Whoops I forgot to force push..

PR #122, opened 2020-09-14
✨(demo) force user.is_superuser on login when they have is_staff True

lunika:
### Purpose

OpenEdx hawthorn doesn't send in the oauth response if the user is a
superuser, only if he is a staff member. For the demo we decided to set
a user as superuser if he is staff. For this we added a function in the
authentication pipeline executed once we are sure a user exists and
check if he is a staff member. If it's the case then we set the user as
a superuser and we save it.

### Proposal

- [x] force user.is_superuser on login when they have is_staff True

### DIFF
--- a/sites/demo/CHANGELOG.md
+++ b/sites/demo/CHANGELOG.md
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Add step in the social auth connection pipeline to force user
+  having is_staff set to true to also have is_superuser set to true.
+
 ## [2.0.0-beta.14.5] - 2020-09-08
 
 ### Fixed
--- a/sites/demo/src/backend/demo/social/pipeline/user.py
+++ b/sites/demo/src/backend/demo/social/pipeline/user.py
@@ -1,0 +1,10 @@
+"""Pipeline used only for the demo to add all permissions to a staff user."""
+
+
+def set_super_user(strategy, details, backend, user=None, *args, **kwargs):
+    """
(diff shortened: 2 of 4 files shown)


sites/demo/src/backend/demo/social/pipeline/user.py
```diff
@@ -0,0 +1,8 @@
+"""Pipeline used only for the demo to add all persmission to a staff user."""
+
+
+def set_super_user(strategy, details, backend, user=None, *args, **kwargs):
+    """set user.is_superuser to True is there is_staff in the details and user exists."""
```
sampaccoud, line 5:
```suggestion
    """set user.is_superuser to True if there is_staff is set in the details and the user exists."""
```

sites/demo/src/backend/demo/social/pipeline/user.py
```diff
@@ -0,0 +1,8 @@
+"""Pipeline used only for the demo to add all persmission to a staff user."""
```
sampaccoud, line 1:
```suggestion
"""Pipeline used only for the demo to add all permissions to a staff user."""
```

sites/demo/CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Function added in the social auth connection pipeline to force user
```
sampaccoud, line 13:
```suggestion
- Add step in the social auth connection pipeline to force user
```

sampaccoud (approved):

PR #123, opened 2020-09-15
Improvement/remove licences block

jbpenrath:
## Purpose
Feedbacks mentionned that licences are not relevant on FUN Campus and FUN Corporate platforms.

## Proposal
- [x] Hide licences block on FUN Corporate
- [x] Hide licences block on FUN Campus

### DIFF
--- a/sites/funcampus/src/backend/templates/courses/cms/course_run_detail.html
+++ b/sites/funcampus/src/backend/templates/courses/cms/course_run_detail.html
@@ -1,0 +1,4 @@
+{% extends "courses/cms/course_run_detail.html" %}
+
+{% comment %}Disable licences part. It does not make sens in FUN campus.{% endcomment %}
+{% block licenses %}{% endblock licenses %}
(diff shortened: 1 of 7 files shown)


sites/funcampus/src/backend/templates/courses/cms/course_run_detail.html
```diff
@@ -0,0 +1,4 @@
+{% extends "courses/cms/course_run_detail.html" %}
```
jbpenrath, line 1: Is it relevant to override `course_run_detail` template?

sites/funcorporate/src/backend/templates/courses/cms/course_run_detail.html
```diff
@@ -0,0 +1,4 @@
+{% extends "courses/cms/course_run_detail.html" %}
```
jbpenrath, line 1: Is it relevant to override `course_run_detail` template?

sites/funcampus/src/backend/templates/courses/cms/course_run_detail.html
sampaccoud: Yes it is. It is a mirror of the course_detail.html template. In a future refactoring I will remove all course run pages because we realized it is not how course runs will be handled IRL. Having to maintain a page for each course run is painful and useless... This will also remove the complexity that you are pointing here...

sites/funcorporate/src/backend/templates/courses/cms/course_run_detail.html
sampaccoud: ditto ;-)

sampaccoud (approved):

PR #129, opened 2020-09-17
Funcorporate: configure contact buttons, fix categories filter and customize translations

sampaccoud:
### Changed

- Customize wording for "What you will learn" using translations

### Fixed

- Filter of "training domains", not on "categories" which is the root node
- Rename contact buttons in header and course page and point to Zendesk


sites/funcorporate/src/backend/funcorporate/settings.py
```diff
@@ -327,13 +327,13 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
         (
             "richie.apps.search.filter_definitions.IndexableMPTTFilterDefinition",
             {
-                "human_name": _("Subjects"),
+                "human_name": _("Training domains"),
```
jbpenrath, line 330: After bootstrap, this key was not translated but i18n files was up to date. I retyped manualy this key and `make i18n` again to see the translated message 🤷‍♂️

sites/funcorporate/src/backend/funcorporate/settings.py
```diff
@@ -327,13 +327,13 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
...
                 "is_autocompletable": True,
                 "is_searchable": True,
                 "min_doc_count": 0,
-                "name": "subjects",
+                "name": "domains",
```
jbpenrath, line 334: Is it normal that with demo site data, this new filter is empty?

sites/funcorporate/src/backend/funcorporate/settings.py
sampaccoud: Maybe it has to do with the fact that this is parsed upon container startup.

sites/funcorporate/src/backend/funcorporate/settings.py
sampaccoud: Yes. It requires a category page with a reverse_id of "domains". I submitted a PR on richie to improve demo site on customized sites and will look into it once we have released the next version of richie.

jbpenrath (approved):

sites/funcorporate/src/backend/funcorporate/settings.py
sampaccoud: Ok got it. It's because the compilemessages command is run in the dockerfile but in docker-compose, we override /app with our local code. So in your case, the .mo file (which is git ignored) was not present at the moment the container started. We should add a `make i18n` command in the bootstrap before `make run`.

jbpenrath: Thanks for explanation

PR #130, opened 2020-09-23
✨(funcampus) add link to dashboard as a button in the header

sampaccoud:
### Purpose

The fun-campus.fr site has no SSO configured. We should still provide a link to the LMS for students who may be subscribed on
a course by their teacher (invitation only).

### Proposal

Add a link to the LMS dashboard as a button in the header.


jbpenrath (approved):
GTM

PR #133, opened 2020-09-29
✨(funcampus) add monkey patch to limit the browser cache TTL

madmatah:
## Purpose

We identified several issues with the caching mechanism.

1. The browser cache TTL reflects the server cache TTL and leads to cache invalidation issues when this TTL is high.
2. Two caching layers are doing the same job : django cache middlewares and djange CMS page cache.

In addition to that, we changed our user management strategy. Most of our users will not have an account on richie. As a
result, the monkey patch that we applied to enable cache to non-staff users is now useless.

## Proposal

- [x] disable cms page cache for non-staff users
- [x] disable django cache middlewares
- [x] add monkey patch to limit the browser cache TTL


sites/funcampus/CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Add monkey patch to limit the browser cache TTL
+
+### Removed
+
+- Remove monkey patch that enable cms page cache for non-staff users
```
sampaccoud, line 17:
```suggestion
- Remove monkey patch that enabled cms page cache for non-staff users
```

sites/funcampus/src/backend/base/admin.py
```diff
@@ -43,12 +43,9 @@ def ajax_upload(request, folder_id=None):
...
 admin.site.register(user_model, CustomUserAdmin)
 
-# Django CMS disables page cache for authenticated users
-# (See https://github.com/divio/django-cms/blob/3.7.4/cms/cache/page.py#L38)
-#
-# In our case, since most of our users will be authenticated and be served the
-# same content, this is a problem. That's why we are monkey patching the
-# function `cms.cache.page.set_page_cache`.
-# All non-staff users will benefit from page cache.
+# By default, Django CMS send cached response with a Cache-control: max-age
```
sampaccoud, line 46:
```suggestion
# By default, Django CMS sends cached responses with a Cache-control: max-age
```

sites/funcampus/src/backend/funcampus/settings.py
```diff
@@ -468,6 +466,17 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
         {"menus": 3600, "content": 86400, "permissions": 86400}
     )
 
+    # By default, Django CMS send cached response with a Cache-control: max-age
```
sampaccoud, line 469:
```suggestion
    # By default, Django CMS sends cached responses with a Cache-control: max-age
```

sites/funcampus/src/backend/funcampus/settings.py
```diff
@@ -468,6 +466,17 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
...
+    # value that reflects the server cache TTL (CMS_CACHE_DURATIONS["content"])
+    #
+    # The thing is : we can invalidate a server side cache entry, but we cannot
+    # invalidate our client browser cache entries. That's why we want to set a
+    # long TTL on the server side, but a much lower TTL on the browser cache.
+    #
+    # This settings allows to define a maximum value for the max-age header
```
sampaccoud, line 476:
```suggestion
    # This setting allows to define a maximum value for the max-age header
```

sites/funcampus/src/backend/funcampus/settings.py
```diff
@@ -468,6 +466,17 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
...
+    # return by Django CMS views.
```
sampaccoud, line 477:
```suggestion
    # returned by Django CMS views.
```

sites/funcampus/src/backend/base/cache.py
```diff
@@ -2,81 +2,42 @@
...
-            cache.set(
-                _page_cache_key(request),
-                (response.content, response._headers, expires_datetime,),
-                ttl,
-                version=version,
-            )
-            # See note in invalidate_cms_page_cache()
-            _set_cache_version(version)
+    response = django_cms_details(request, *args, **kwargs)
+    # The response return by cms.views.detail is either a TemplateResponse or a BaseHttpResponse
```
sampaccoud, line 34:
```suggestion
    # The response returned by cms.views.detail is either a TemplateResponse or a BaseHttpResponse
```

sites/funcampus/src/backend/base/cache.py
```diff
@@ -2,81 +2,42 @@
...
+    # In the case of a TemplateResponse, the rendering can be delayed and the cache headers not
+    # set yet. That's why we have to add a post rendering-callback to patch the response headers.
+    if hasattr(response, "is_rendered") and not response.is_rendered:
```
sampaccoud, line 37: I'm wondering, since what you need to do is to modify a header on the response, why not do it in a small middleware that we add at the last position of our middlewares!?

madmatah: I considered this middleware solution, but what bothered me is that a middleware acts on all responses.
How to target only the one returned by Django CMS ?

I didn't want to introduce a middleware that could disturb other django applications.

sampaccoud: I would say that the cache approach in Django is anyway across the site so it makes sense to centralize all modifications to a middleware? In our case this site will only be DjangoCMS so if we apply this rule in this repository it seems ok?
I think it's worth avoiding the monkey patch if we can :sweat_smile:

madmatah: I understand your point of view. I'm just not sure about the durability of this statement «this site will only be DjangoCMS» :smile: 

I'll update my PR to do this with a middleware.

sampaccoud:
Closed in favor of PR https://github.com/openfun/richie-site-factory/pull/137 as we decided to use a middleware and added it to richie.

PR #134, opened 2020-09-30
🔖(chore) bump demo to 2.0.0-beta.14.6

lunika:
Added

- Add step in the social auth connection pipeline to force user
  having is_staff set to true to also have is_superuser set to true.


sampaccoud:
we are releasing a new richie version today. Let's wait and save some work.

PR #138, opened 2020-10-08
✨(project) add script to automate tagging

sampaccoud:
## Purpose

Once release commits have landed to master, we need to tag them before pushing the tags to github. This is repetitive and painful
so we should automate it.

## Proposal

Add a `tag` script that:

- looks for commits starting with the 🔖 gitmoji, 
- extracts the site and version from the commit subject
- tags the commit with the corresponding tag of the form {site}-{version}

### DIFF
--- a/bin/tag
+++ b/bin/tag
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# shellcheck source=bin/_config.sh
+source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
+
+function usage(){
+    echo -e "Usage: ${0} [OPTIONS]
+
+Tag all untagged release commits (starting with the 🔖 gitmoji) and
+push the tags to github so the image is built and pushed to DockerHub.
+
+OPTIONS:
+  -h, --help       print this message
+  -c, --commit     dry run if this option is not set
+"
+}
+
+declare -i should_commit=0
+declare regex="🔖\(.*\).*version [0-9\.]*"
+
+# Parse options
+for i in "$@"
+do
+    case $i in
+        -h|--help|help)
+            usage "${0}"
+            exit 0
+            ;;
+        -c|--commit)
+            should_commit=1
+            shift
+            ;;
+        *)
+            shift
+            ;;
+    esac
+done
+
+git fetch origin
+
+# Loop over new release commits that were not tagged yet
+for commit_hash in $(git log --grep="${regex}" --since=2020-10-07 --pretty=format:"%h %d" | grep -v "tag:" | awk '{print $1}')
+do
+    version=$(git show --format="%s" -s "${commit_hash}" | sed -E "s/.*version ([0-9\.]*).*/\1/")
+    site=$(git show --format="%s" -s "${commit_hash}" | sed -E "s/🔖\((.*)\).*/\1/")
+    action=$([[ "${should_commit}" == 1 ]] && echo "Tagging" || echo "Would tag")
+
+    echo -e "${COLOR_INFO}${action} ${commit_hash} with ${site}-${version}${COLOR_RESET}"
+    if [[ "${should_commit}" == 1 ]]; then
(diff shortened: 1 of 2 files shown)


bin/tag
```diff
@@ -0,0 +1,53 @@
...
+            should_commit=1
+            shift
+            ;;
+        *)
+            shift
+            ;;
+    esac
+done
+
+
```
jmaupetit, line 40: You need  to `git fetch origin` first in case you didn't fetch tags created and pushed by one of our coworkers.

bin/tag
```diff
@@ -0,0 +1,53 @@
...
+for commit_hash in $(git log --grep="${regex}" --since=2020-10-07 --pretty=format:"%h %d" | grep -v "tag:" | awk '{print $1}')
+do
+    version=$(git show --format="%s" -s "${commit_hash}" | sed -E "s/.*version ([0-9\.]*).*/\1/")
+    site=$(git show --format="%s" -s "${commit_hash}" | sed -E "s/🔖\((.*)\).*/\1/")
+    action=$([[ "${should_commit}" == 1 ]] && echo "Tagging" || echo "Would tag")
+
+    echo -e "${COLOR_INFO}${action} ${commit_hash} with ${site}-${version}${COLOR_RESET}"
+    if [[ "${should_commit}" == 1 ]]; then
+        git tag "${site}-${version}" "${commit_hash}"
+        git push origin --tag
```
jmaupetit, line 51:
```suggestion
        git push origin --tags
```

jmaupetit (commented):
Another approach would have been to use GitHub's API as we did for https://github.com/openfun/learninglocker-docker (see the [`bin/release` script](https://github.com/openfun/learninglocker-docker/blob/master/bin/release))

sampaccoud:
@jmaupetit does you other approach take care of creating releases in github? This is one part I'm interested in adding?

jmaupetit:
> @jmaupetit does you other approach take care of creating releases in github? This is one part I'm interested in adding?

Unfortunately not. Check this https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#create-a-release

Combined with our script mentioned previously, you should be able to do so.

bin/tag
sampaccoud: nice catch :pray:

bin/tag
sampaccoud: :+1:

sampaccoud:
@jmaupetit if it does not handle github releases I prefer to keep the current approach as it is simpler.

To automate github releases, I'm starting to think about a github action that watches the changelog and generates the release from it... The advantage is that this action could be added without any modification to all our repositories!

bin/tag
```diff
@@ -0,0 +1,54 @@
...
+# Loop over new release commits that were not tagged yet
+for commit_hash in $(git log --grep="${regex}" --since=2020-10-07 --pretty=format:"%h %d" | grep -v "tag:" | awk '{print $1}')
+do
+    version=$(git show --format="%s" -s "${commit_hash}" | sed -E "s/.*version ([0-9\.]*).*/\1/")
+    site=$(git show --format="%s" -s "${commit_hash}" | sed -E "s/🔖\((.*)\).*/\1/")
+    action=$([[ "${should_commit}" == 1 ]] && echo "Tagging" || echo "Would tag")
+
+    echo -e "${COLOR_INFO}${action} ${commit_hash} with ${site}-${version}${COLOR_RESET}"
+    if [[ "${should_commit}" == 1 ]]; then
+        git fetch origin
```
jmaupetit, line 50: I would rather do this before the loop.

jmaupetit (commented):
Ok to try github actions for our releases (theoretically)!

bin/tag
sampaccoud: Makes sense! Done.

jmaupetit (approved):

sampaccoud:
@jmaupetit tk tk

PR #143, opened 2020-10-14
🔧(funcorporate) customize max number of plugins in course description

sampaccoud:
## Purpose

Max number of plugins in course description needs to be temporarily relaxed because we have legacy content that is very long and that we can't rewrite in the short term.

## Proposal

- [x] Introduce a new setting and make a deep merge with the existing dictionary for placeholders configuration.
- [x] Bump max number of plugins in the course description to 2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -347,73 +347,73 @@ workflows:
                 name: no-change-demo
     funcampus:
         jobs:
+            - no-change:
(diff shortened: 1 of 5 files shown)


madmatah (approved):
LGTM :ok_hand:

jbpenrath (approved):

PR #148, opened 2020-10-25
Funcorporate proxy cdn to www

sampaccoud:
## Purpose

We moved the richie site to the main domain https://www.fun-corporate.fr. As a result, the CDN must update its origin.

### Proposal

Point CDN to www.fun-corporate.fr


jbpenrath (approved):
GTM!

PR #154, opened 2020-11-04
Add media files expiration lifecyle rule for demo site

sampaccoud:
## Purpose

The demo site is reset every night. This generates a huge amount of media files because new files are added at each reset! 

## Proposal

Adding an expiration lifecycle rule on the bucket is an easy way to keep it under control. I made it generic and configurable for each site.

### DIFF
--- a/aws/s3.tf
+++ b/aws/s3.tf
@@ -12,7 +12,18 @@ resource "aws_s3_bucket" "richie_media" {
   }
 
   versioning {
-    enabled = true
+    enabled = var.media_expiration > 0 ? false : true
+  }
+
+  dynamic "lifecycle_rule" {
+    for_each = var.media_expiration > 0 ? [1] : []
+    content {
+      id      = "expiration"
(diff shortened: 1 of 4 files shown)


aws/s3.tf
```diff
@@ -12,7 +12,18 @@ resource "aws_s3_bucket" "richie_media" {
   }
 
   versioning {
-    enabled = true
+    enabled = var.media_expiration > 0 ? false : true
+  }
+
+  dynamic "lifecycle_rule" {
+    for_each = var.media_expiration > 0 ? [1] : []
```
lunika, line 19: I don't see where you are using it. Is it a hack to not declare this block if `media_expiration` equal `0` ?

sampaccoud: it's not really a hack but the best way I found to do a conditional blocks in Terraform.
You can consider it as a "if"...

lunika (approved):

PR #162, opened 2020-11-12
🚨(bin/release) should_commit does not work on Mac OS

jbpenrath:
## Purpose
`head -n -1` did not work on Mac OS so it is impossible to commit automatically changes.

## Proposal
- [x] use `sed \$d` instead of `head -n -1` to get all `$changelog` content but the last line.


sampaccoud (approved):
Still working by me :+1:

PR #163, opened 2020-11-12
Fix media backend following upgrade to DjangoCMS 3.8 and release fix releases

sampaccoud:
### Purpose

The latest upgrades have broken the media file storage backend.

### Proposal

- [x] Force the media storage backend location to "media".
- [x] Upgrade `django-storage` to its latest version

:information_source: I took the opportunity to fix the docker build which was very slow. In fact the reason was that the `node_modules` folder was no more targeted by the `.dockerignore` file since the refactoring to a multi-site factory.


sites/cnfpt/src/backend/base/storage.py
```diff
@@ -58,4 +58,4 @@ class MediaStorage(S3Boto3Storage):
     bucket_name = getattr(settings, "AWS_MEDIA_BUCKET_NAME", None)
     custom_domain = getattr(settings, "CDN_DOMAIN", None)
     file_overwrite = False
-    location = settings.MEDIA_URL
```
jbpenrath, line 61: Why not just update `MEDIA_URL` in settings.py?

jbpenrath (approved):

jbpenrath:
Ignore `node_modules` really speed ud building 🎉

sites/cnfpt/src/backend/base/storage.py
sampaccoud: I did not manage to do it with "location" as a class property because it was evaluated too early. 
I reverted to the AWS_LOCATION setting and it works. :+1:

sampaccoud:
@jbpenrath I added a fix for the demo site

sites/funcampus/src/backend/funcampus/settings.py
```diff
@@ -539,6 +539,38 @@ class Development(Base):
         }
     }
 
+    RICHIE_DEMO_NB_OBJECTS = {"licences": 0}
```
jbpenrath, line 542: Your commit mention demo site as scope but you edit file in funcampus?

sampaccoud: True. It's the demo of funcampus. Changed it.

PR #164, opened 2020-11-15
Pin django to 3.1.1

sampaccoud:
## Purpose

A CSS change in [Django 3.1.2]() broke the layout of DjangoCMS's admin pages. 

image

## Proposal

While waiting for a fix in DjangoCMS, we can pin Django to fix it.

I opened an issue on the DjangoCMS project: https://github.com/django-cms/django-cms/issues/6944


sites/cnfpt/requirements/base.txt
```diff
@@ -1,4 +1,6 @@
 boto3==1.9.96
+# /admin/cms/page is broken with Django>=3.1.2
+Django<=3.1.1
```
jbpenrath, line 3: `Django==3.1.1` ?

jbpenrath (approved):
Tested on CNFPT, admin views seems ok.

sites/cnfpt/requirements/base.txt
sampaccoud: yeah it's the same :man_shrugging:

PR #170, opened 2020-12-09
Include version to cms cache prefix + set english as primary language for demo site + issue fix releases

sampaccoud:
### Purpose

When deploying a new version of the app, the cache was is reset. This can lead to bugs if the cached html and the new javascript bundle are not incompatible. 

### Proposal

Including the version of the app in the CMS cache prefix makes sure that each version of the app handles its cache independently.


jbpenrath (approved):
LGTM

PR #171, opened 2020-12-10
[funmooc] activate cache

sampaccoud:
### Fixed

- Configure cache durations to make cache effective in production


jbpenrath (approved):
😳

PR #179, opened 2020-12-14
🌐(cnfpt) replace MOOCs by MOOC

jbpenrath:
With a few exceptions, an abbreviation does not take the plural form.


sampaccoud:
Nice. 
Maybe you can add the fix release to this PR to save time :man_shrugging:

sampaccoud (approved):

PR #190, opened 2021-01-06
Fallback cache when redis cache is down

jbpenrath:
### Purpose

When Redis cache is out of service, an exception is catched and the platform is down which is awkward.

### Proposal

- [x] Create a custom cache `RedisCacheWithFallback` which is able to use a fallback cache in case redis is down.

### DIFF
--- a/.circleci/src/jobs/@backend.yml
+++ b/.circleci/src/jobs/@backend.yml
@@ -43,7 +43,7 @@ lint-back:
         command: ~/.local/bin/flake8
     - run:
         name: Lint code with isort
-        command: ~/.local/bin/isort --recursive --check-only .
+        command: ~/.local/bin/isort --check-only .
     - run:
         name: Lint code with black
         command: ~/.local/bin/black .
--- a/sites/cnfpt/src/backend/base/cache.py
+++ b/sites/cnfpt/src/backend/base/cache.py
@@ -1,0 +1,184 @@
+"""
+    RedisCache with a fallback cache to prevent denial of service if Redis is down
+    Freely inspired by django-cache-fallback
+
+    Credits:
+    - https://github.com/Kub-AT/django-cache-fallback/
+"""
+
+import logging
+
+from django.conf import settings
+from django.core.cache import caches
+from django.core.cache.backends.base import BaseCache
+
+from cms.cache import CMS_PAGE_CACHE_VERSION_KEY
+from django_redis.cache import RedisCache
+
+from .utils import throttle
+
+FALLBACK_CACHE_INVALIDATION_INTERVAL = 60  # seconds
+DJANGO_REDIS_LOGGER = getattr(settings, "DJANGO_REDIS_LOGGER", __name__)
+logger = logging.getLogger(DJANGO_REDIS_LOGGER)
+
+
+class RedisCacheWithFallback(BaseCache):
+    """
+    BaseCache object with a redis_cache used as main cache
+    and the "fallback" aliased cache which takes over
+    in case redis_cache is down.
+    """
+
+    def __init__(self, server, params):
+        """
+        Instantiate the Redis Cache with server and params
+        and retrieve the cache with alias "fallback"
+        """
+        super().__init__(params)
+        self._redis_cache = RedisCache(server, params)
+        self._fallback_cache = caches["fallback"]
+
+    def _call_with_fallback(self, method, *args, **kwargs):
+        """
+        Try first to exec provided method through Redis cache instance,
+        in case of success, invalidate the fallback cache so it is clean and
+        ready for next failure,
+        in case of failure, logger reports the exception and
+        the fallback cache takes over.
+        """
+        try:
+            next_cache_state = self._call_redis_cache(method, args, kwargs)
+        except Exception as e:
+            logger.warning("[DEGRADED CACHE MODE] - Switch to fallback cache")
+            logger.exception(e)
+            return self._call_fallback_cache(method, args, kwargs)
+        else:
+            self._invalidate_fallback_cache()
+            return next_cache_state
+
+    def _call_redis_cache(self, method, args, kwargs):
+        """
+        Exec the provided method through the redis cache instance
+        """
+        return getattr(self._redis_cache, method)(*args, **kwargs)
+
+    def _call_fallback_cache(self, method, args, kwargs):
+        """
+        Exec the provided method through the fallback cache instance
+        """
+        return getattr(self._fallback_cache, method)(*args, **kwargs)
+
+    @throttle(FALLBACK_CACHE_INVALIDATION_INTERVAL)
+    def _invalidate_fallback_cache(self):
+        """
+        Invalidate cms page cache in the fallback cache.
+        """
+        self._fallback_cache.delete(CMS_PAGE_CACHE_VERSION_KEY)
+
+    def get_backend_timeout(self, *args, **kwargs):
+        """
+        Pass get_backend_timeout cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("get_backend_timeout", *args, **kwargs)
+
+    def make_key(self, *args, **kwargs):
+        """
+        Pass make_key cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("make_key", *args, **kwargs)
+
+    def add(self, *args, **kwargs):
+        """
+        Pass add cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("add", *args, **kwargs)
+
+    def get(self, *args, **kwargs):
+        """
+        Pass get cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("get", *args, **kwargs)
+
+    def set(self, *args, **kwargs):
+        """
+        Pass set cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("set", *args, **kwargs)
+
+    def touch(self, *args, **kwargs):
+        """
+        Pass touch cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("touch", *args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        """
+        Pass delete cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("delete", *args, **kwargs)
+
+    def get_many(self, *args, **kwargs):
+        """
+        Pass get_many cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("get_many", *args, **kwargs)
+
+    def get_or_set(self, *args, **kwargs):
+        """
+        Pass get_or_set cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("get_or_set", *args, **kwargs)
+
+    def has_key(self, *args, **kwargs):
+        """
+        Pass has_key cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("has_key", *args, **kwargs)
+
+    def incr(self, *args, **kwargs):
+        """
+        Pass incr cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("incr", *args, **kwargs)
+
+    def decr(self, *args, **kwargs):
+        """
+        Pass decr cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("decr", *args, **kwargs)
+
+    def set_many(self, *args, **kwargs):
+        """
+        Pass set_many cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("set_many", *args, **kwargs)
+
+    def delete_many(self, *args, **kwargs):
+        """
+        Pass delete_many cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("delete_many", *args, **kwargs)
+
+    def clear(self):
+        """
+        Pass clear cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("clear")
+
+    def validate_key(self, *args, **kwargs):
+        """
+        Pass validate_key cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("validate_key", *args, **kwargs)
+
+    def incr_version(self, *args, **kwargs):
+        """
+        Pass incr_version cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("incr_version", *args, **kwargs)
+
+    def decr_version(self, *args, **kwargs):
+        """
+        Pass decr_version cache method to _call_with_fallback
+        """
+        return self._call_with_fallback("decr_version", *args, **kwargs)
--- a/sites/cnfpt/src/backend/base/tests/test_cache.py
+++ b/sites/cnfpt/src/backend/base/tests/test_cache.py
@@ -1,0 +1,159 @@
+"""Test funmooc cache plugin"""
+
+import datetime
+from unittest import mock
+
+from django.core.cache.backends.dummy import DummyCache
+from django.test import TestCase, override_settings
+
+from django_redis.cache import RedisCache
+
+from base.cache import RedisCacheWithFallback
+
+
+class RedisCacheWithFallbackTestCase(TestCase):
+    """
+    Test suite for the RedisCacheWithFallback
+
+    Credits:
+    - https://github.com/Kub-AT/django-cache-fallback
+    """
+
+    @override_settings(
+        CACHES={
+            "default": {
+                "BACKEND": "base.cache.RedisCacheWithFallback",
+                "LOCATION": "mymaster/redis-sentinel:26379,redis-sentinel:26379/0",
+                "OPTIONS": {
+                    "CLIENT_CLASS": "richie.apps.core.cache.SentinelClient",
+                },
+            },
+            "fallback": {
+                "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+            },
+        }
+    )
+    def test_client(self):
+        """Test class instance of caches"""
+        client = RedisCacheWithFallback(None, {})
+        self.assertIs(type(client), RedisCacheWithFallback)
+        self.assertIs(type(client._redis_cache), RedisCache)
+        self.assertIs(type(client._fallback_cache), DummyCache)
+
+    @mock.patch.object(RedisCacheWithFallback, "_call_fallback_cache")
+    @mock.patch.object(RedisCacheWithFallback, "_call_redis_cache")
+    def test_get_redis_cache(self, redis_cache_mock, fallback_cache_mock):
+        """Test that redis_cache is used by default."""
+        client = RedisCacheWithFallback(None, {})
+        client.get("irrelevent")
+
+        redis_cache_mock.assert_called_once()
+        fallback_cache_mock.assert_not_called()
+
+    @mock.patch("base.cache.logger")
+    @mock.patch.object(RedisCacheWithFallback, "_call_fallback_cache")
+    @mock.patch.object(RedisCacheWithFallback, "_call_redis_cache")
+    def test_get_fallback_cache(
+        self, redis_cache_mock, fallback_cache_mock, logger_mock
+    ):
+        """
+        Test case when redis_cache raises an exception,
+        logger logs the exception then fallback_cache takes over
+        """
+        client = RedisCacheWithFallback(None, {})
+        client.get("irrelevent")
+
+        redis_cache_mock.assert_called_once()
+        fallback_cache_mock.assert_not_called()
+
+        redis_cache_mock.side_effect = Exception()
+        client.get("irrelevent")
+
+        logger_mock.warning.assert_called_with(
+            "[DEGRADED CACHE MODE] - Switch to fallback cache"
+        )
+        logger_mock.exception.assert_called_once()
+        fallback_cache_mock.assert_called_once()
+
+    @override_settings(
+        CACHES={
+            "default": {
+                "BACKEND": "base.cache.RedisCacheWithFallback",
+                "LOCATION": "mymaster/redis-sentinel:26379,redis-sentinel:26379/0",
+                "OPTIONS": {
+                    "CLIENT_CLASS": "richie.apps.core.cache.SentinelClient",
+                },
+            },
+            "fallback": {
+                "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+            },
+        }
+    )
+    @mock.patch.object(DummyCache, "delete")
+    @mock.patch.object(RedisCacheWithFallback, "_call_redis_cache")
+    def test_invalidate_fallback_cache(self, redis_cache_mock, delete_mock):
+        """
+        Test that fallback_cache is invalidated every 60 seconds
+        when Redis is up
+        """
+        now = datetime.datetime.now()
+
+        client = RedisCacheWithFallback(None, {})
+        client.get("round_0")
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        # A second cache access should not delete the fallback cache again
+        client.get("round_1")
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_not_called()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        # Another call to Redis cache 120 seconds later
+        # should delete the fallback cache again
+        read_time = now + datetime.timedelta(seconds=120)
+        with mock.patch("base.utils.datetime") as mock_datetime:
+            mock_datetime.now.return_value = read_time
+            client.get("round_2")
+
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_called_once()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        # Another call to Redis cache 30 seconds later (<1 minute)
+        # should not delete the fallback cache again
+        read_time += datetime.timedelta(seconds=30)
+        with mock.patch("base.utils.datetime") as mock_datetime:
+            mock_datetime.now.return_value = read_time
+            client.get("round_3")
+
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_not_called()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        # Another call to Redis cache exactly 1 minute after
+        # the latest call should not delete the fallback cache again
+        read_time += datetime.timedelta(seconds=30)
+        with mock.patch("base.utils.datetime") as mock_datetime:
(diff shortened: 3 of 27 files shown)


sites/funmooc/CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Use a custom RedisCacheWithFallback cache to prevent deny of service
+  when Redis is downed
```
sampaccoud, line 14:
```suggestion
- Use a custom RedisCacheWithFallback cache backend to maintain site up
  when Redis is down
```

sites/funmooc/src/backend/base/cache.py
```diff
@@ -0,0 +1,111 @@
+"""
+  RedisCache with a fallback cache to prevent deny of service if Redis is downed
```
sampaccoud, line 2:
```suggestion
  RedisCache with a fallback cache to prevent denial of service if Redis is down
```

sites/funmooc/src/backend/funmooc/settings.py
```diff
@@ -553,7 +556,10 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
                     environ_prefix=None,
                 ),
                 "OPTIONS": values.DictValue(
-                    {"CLIENT_CLASS": "richie.apps.core.cache.SentinelClient"},
+                    {
+                        "CLIENT_CLASS": "richie.apps.core.cache.SentinelClient",
+                        "FALLBACK_CLASS": "django.core.cache.backends.locmem.LocMemCache",
```
sampaccoud, line 561: the setting is called `BACKEND` and not `BACKEND_CLASS`, so maybe call it `FALLBACK_BACKEND` to differentiate from the the sentinel client class:
```suggestion
                        "FALLBACK_BACKEND": "django.core.cache.backends.locmem.LocMemCache",
```

sites/funmooc/src/backend/funmooc/settings.py
```diff
@@ -538,12 +538,15 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
         default="richie", environ_name="RICHIE_ES_INDICES_PREFIX", environ_prefix=None
     )
 
+    # DJANGO_REDIS_IGNORE_EXCEPTIONS = True
+    # DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = True
```
sampaccoud, line 542: to be removed?

sites/funmooc/src/backend/base/cache.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, server, params):
+        super().__init__(params)
+        options = params.get("OPTIONS", {})
+        self._redis_cache = RedisCache(server, params)
+        self._fallback_cache = self.initialize_fallback_cache(
+            options.get(
+                "FALLBACK_CLASS", "django.core.cache.backends.locmem.LocMemCache"
+            )
+        )
+        self.log = logging.getLogger(DJANGO_REDIS_LOGGER)
```
sampaccoud, line 30: just use your `logger` above and remove this?

sites/funmooc/src/backend/base/cache.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @staticmethod
+    def initialize_fallback_cache(backend, **kwargs):
+        fallback_cache = cache._create_cache(backend, **kwargs)
```
sampaccoud, line 34: add docstrings to every functions

sites/funmooc/src/backend/base/cache.py
```diff
@@ -0,0 +1,111 @@
...
+        return getattr(self._redis_cache, method)(*args, **kwargs)
+
+    def _call_fallback_cache(self, method, args, kwargs):
+        return getattr(self._fallback_cache, method)(*args, **kwargs)
+
+    def get_backend_timeout(self, *args, **kwargs):
+        return self._call_with_fallback("get_backend_timeout", *args, **kwargs)
+
+    def make_key(self, *args, **kwargs):
+        return self._call_with_fallback("add", *args, **kwargs)
```
sampaccoud, line 60: should it be make_key?
```suggestion
        return self._call_with_fallback("make_key", *args, **kwargs)
```

sites/funmooc/src/backend/base/cache.py
```diff
@@ -0,0 +1,111 @@
...
+        return self._call_with_fallback("set", *args, **kwargs)
+
+    def touch(self, *args, **kwargs):
+        return self._call_with_fallback("touch", *args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        return self._call_with_fallback("delete", *args, **kwargs)
+
+    def get_many(self, *args, **kwargs):
+        return self._call_with_fallback("touch", *args, **kwargs)
```
sampaccoud, line 78: `get_many` vs `touch`?

sites/funmooc/src/backend/base/cache.py
jbpenrath: Copy pasta failed

sites/funmooc/src/backend/base/cache.py
jbpenrath: Ditto

.circleci/src/jobs/@backend.yml
```diff
@@ -43,7 +43,7 @@ lint-back:
         command: ~/.local/bin/flake8
     - run:
         name: Lint code with isort
-        command: ~/.local/bin/isort --recursive --check-only .
```
sampaccoud, line 46: so are you sure that it is still recursively checking the whole `.` directory?

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        Exec the provided method through the fallback cache instance
+        """
+        return getattr(self._fallback_cache, method)(*args, **kwargs)
+
+    def _invalidate_fallback_cache_if_needed(self):
+        """
+        Check if fallback has been used, if yes invalidate its cms page cache.
+        """
+        if self._fallback_cache.get(INVALIDATION_FALLBACK_CACHE_KEY) is True:
```
sampaccoud, line 93: I don't see the interest of checking "if needed". I think you should just invalidate it whatever the situation as long as it is requested.

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,209 @@
...
+import logging
+
+from django.conf import settings
+from django.core import cache, signals
+from django.core.cache.backends.base import BaseCache
+
+from cms.utils.conf import get_cms_setting
+from django_redis.cache import RedisCache
+
+CMS_PAGE_CACHE_VERSION_KEY = get_cms_setting("CACHE_PREFIX") + "_PAGE_CACHE_VERSION"
```
sampaccoud, line 18:
```suggestion
from cms.cache import CMS_PAGE_CACHE_VERSION_KEY
```

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,209 @@
...
+    def _invalidate_fallback_cache_if_needed(self):
+        """
+        Check if fallback has been used, if yes invalidate its cms page cache.
+        """
+        if self._fallback_cache.get(INVALIDATION_FALLBACK_CACHE_KEY) is True:
+            version = self._fallback_cache.get(CMS_PAGE_CACHE_VERSION_KEY)
+            if version:
+                self._fallback_cache.set(
+                    CMS_PAGE_CACHE_VERSION_KEY,
+                    version,
```
sampaccoud, line 98: should be version + 1?

sites/cnfpt/src/backend/base/cache.py
jbpenrath: Just limit cache operation cost when it is not needed.

sampaccoud: but you have to make a get so it's counter productive? As this is local memory cache I don't think it is worth it anyway.

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,209 @@
...
+                    get_cms_setting("CACHE_DURATIONS")["content"],
+                )
```
sampaccoud, line 100: Instead of doing all this, maybe you could just you "incr"? 
https://redis.io/commands/incr

sampaccoud: > Increments the number stored at key by one. If the key does not exist, it is set to 0 before performing the operation.

.circleci/src/jobs/@backend.yml
jbpenrath: According to the doc, yes.
https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0/#-recursive-or-rc

> In 5.0.0 directories are automatically traversed for all Python files

sites/cnfpt/src/backend/base/cache.py
jbpenrath: `incr` method do not accept a timeout parameter, so with this method I'm not sure that key lifetime is renewed.

https://github.com/django/django/blob/master/django/core/cache/backends/locmem.py#L69

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,192 @@
...
+            )
+        )
+
+    @staticmethod
+    def initialize_fallback_cache(backend, **kwargs):
+        """
+        Initialize fallback cache with the provided backend then return the
+        cache instance freshly created
+        """
+        fallback_cache = cache._create_cache(backend, **kwargs)
```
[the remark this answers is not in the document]
jbpenrath: I discovered yesterday that `cache._create_cache` used in `initialize_fallback_cache` has been recently removed from django (https://github.com/django/django/commit/148702e7258eafb27d5488b33a723c87d898e22b) (may be not yet released?)

So to use a more "standard" way and simplify tweaking of fallback cache, now fallback cache must be defined in settings as a second cache in `CACHES` then we retrieve this cache from `caches['fallback']`.

sites/cnfpt/src/backend/base/cache.py
jbpenrath: By removing this check, we make 2 extra cache accesses per redis request instead of 1 in case invalidation is not required and 4 when invalidation is needed.

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,184 @@
...
+
+DJANGO_REDIS_LOGGER = getattr(settings, "DJANGO_REDIS_LOGGER", __name__)
+logger = logging.getLogger(DJANGO_REDIS_LOGGER)
+
+
+class RedisCacheWithFallback(BaseCache):
+    """
+    BaseCache object with a redis_cache used as main cache
+    and used the "fallback" aliased cachewhich takes over
+    in case redis_cache is down.
```
sampaccoud, line 27:
```suggestion
    and the "fallback" aliased cache which takes over
    in case redis_cache is down.
```

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,184 @@
...
+        the fallback cache takes over.
+        """
+        try:
+            next_cache_state = self._call_redis_cache(method, args, kwargs)
+            self._invalidate_fallback_cache()
+            return next_cache_state
+        except Exception as e:
+            logger.warning("[DEGRADED CACHE MODE] - Switch to fallback cache")
+            logger.exception(e)
+            return self._call_fallback_cache(method, args, kwargs)
```
sampaccoud, line 52:
```suggestion
        try:
            next_cache_state = self._call_redis_cache(method, args, kwargs)
        except Exception as e:
            logger.warning("[DEGRADED CACHE MODE] - Switch to fallback cache")
            logger.exception(e)
            return self._call_fallback_cache(method, args, kwargs)
        else:
            self._invalidate_fallback_cache()
            return next_cache_state
```

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def _call_fallback_cache(self, method, args, kwargs):
+        """
+        Exec the provided method through the fallback cache instance
+        """
+        return getattr(self._fallback_cache, method)(*args, **kwargs)
+
+    def _invalidate_fallback_cache(self):
+        """
+        From fallback cache, invalidate cms page cache.
```
sampaccoud, line 68:
```suggestion
        Invalidate cms page cache in the fallback cache.
```

sites/cnfpt/src/backend/base/cache.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        version = self._fallback_cache.get(CMS_PAGE_CACHE_VERSION_KEY)
+        if version:
+            self._fallback_cache.set(
+                CMS_PAGE_CACHE_VERSION_KEY,
+                version + 1,
+                get_cms_setting("CACHE_DURATIONS")["content"],
+            )
```
sampaccoud, line 76: I'm thinking : this is going to increment the version VERY high ^^ Why not delete the value completed since we are not trying to bump the cache version but to reset it completely?

sites/funcampus/src/backend/base/cache.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def _call_with_fallback(self, method, *args, **kwargs):
+        """
+        Try first to exec provided method through Redis cache instance,
+        in case of failure, logger reports the exception and
+        the fallback cache takes over.
+        """
+        try:
+            next_cache_state = self._call_redis_cache(method, args, kwargs)
+            self._invalidate_fallback_cache()
```
madmatah, line 47: I'm not comfortable with the fact that this function is called systematically.

One might think that it's okay because the fallback cache is in-memory and operations are fast. But the fallback cache backend is configurable and could be a remote cache, which involves more I/O expensive operations. It could reduce performances even in a healthy situation.

An improvement to solve this problem could be to store the «last successfully used cache backend». And invalidate the fallback cache only if it changes ?

jbpenrath: Yes, this is the big topic of this PR with @sampaccoud.

From my undestanding, it's tricky to share mutable class attribute between class instance. So we have to store this information directly in the fallback cache.

In this case, we have to get the value in the cache then delete the key if needed (2 operations). So @sampaccoud suggests to just make one operation all time : delete the CMS Page cache key.

sites/cnfpt/src/backend/base/cache.py
jbpenrath: You are right! In this way, we make just one operation 🎉

sites/funcampus/src/backend/base/cache.py
sampaccoud: @madmatah makes a good point on the fact that this cache could be somewhere else (although not in this repository which is down the food chain...). So if we go back to which we want to clean this fallback cache in the first place... ie avoid going back to an old version of the site following a switch back and forth between the Redis and the fallback cache... the benefits may not outweigh the costs... The fallback cache is a degraded situation so I think we should take the risk of displaying an old version of the page. Note that the duration of the fallback cache can be set to something low like 10 minutes?

sampaccoud: Or clean but only on round minutes? 😅

madmatah: > From my undestanding, it's tricky to share mutable class attribute between class instance. So we have to store this information directly in the fallback cache.

My bad, I thought there was only one instance of the `RedisCacheWithFallback` class.
My suggestion is not the right way to do it, then :)

jbpenrath: Yes @sampaccoud, let's debounce invalidation each minutes.

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,83 @@
...
+                },
+            },
+            "fallback": {
+                "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+            },
+        }
+    )
+    def test_client(self):
+        """Test class instance of caches"""
+        client = get_client()
```
sampaccoud, line 45: I find this confusing (with self.client). Imo it's clearer to directly write this:
```suggestion
        cache = RedisCacheWithFallback(None, kwargs)
```

sites/cnfpt/src/backend/base/utils.py
```diff
@@ -0,0 +1,36 @@
...
+    def __call__(self, callback):
+        """
+        Process `time_since_last_call`,
+        if it is greater than `throttle_interval`, callback is executed.
+        """
+
+        @wraps(callback)
+        def wrapper(*args, **kwargs):
+            now = datetime.now()
+            time_since_last_call = now - self.time_of_last_call
```
sampaccoud, line 30:
```suggestion
            elapsed_since_last_call = now - self.time_of_last_call
```

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,83 @@
...
+    return client._fallback_cache.get("FALLBACK_CACHE__INVALIDATION_REQUIRED")
+
+
+class RedisCacheWithFallbackTestCase(TestCase):
+    """
+    Test suite for the RedisCacheWithFallback
+
+    Credits:
+    - https://github.com/Kub-AT/django-cache-fallback
+    """
```
sampaccoud, line 27: It might be worth testing the throttle function.  Can be done by mocking the `now` method? https://github.com/openfun/richie/blob/85e0a6aa3a4bff71cad8520089605009af359d5e/tests/apps/search/test_index_manager.py#L139

sampaccoud (approved):

sites/funmooc/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,109 @@
...
+        CACHES={
+            "fallback": {
+                "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+            },
+        }
+    )
+    @mock.patch.object(RedisCacheWithFallback, "_call_redis_cache")
+    def test_invalidate_fallback_cache(self, redis_cache_mock):
+        """
+        Test that fallback_cache is invalidate each 60 seconds
```
sampaccoud, line 88:
```suggestion
        Test that fallback_cache is invalidated every 60 seconds
```

sites/funmooc/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        m = mock.MagicMock()
+        m.delete = mock.Mock(return_value="delete")
+
+        with mock.patch(
+            "django.core.cache.backends.locmem.LocMemCache", return_value=m
+        ) as locmem_mock:
+            client = RedisCacheWithFallback(None, {})
+            client.get("irrelevant")
+            client.get("irrelevant")
```
sampaccoud, line 99: I think you should test the result of each action separately:
```suggestion
            client.get("irrelevant")
            self.assertEqual(redis_cache_mock.call_count, 2)
            self.assertEqual(locmem_mock.delete.call_count, 1)
            redis_cache_mock.reset()
            locmem_mock.reset()
```

sites/funmooc/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Test that fallback_cache is invalidate each 60 seconds
+        when Redis is up
+        """
+        m = mock.MagicMock()
+        m.delete = mock.Mock(return_value="delete")
+
+        with mock.patch(
+            "django.core.cache.backends.locmem.LocMemCache", return_value=m
+        ) as locmem_mock:
```
sampaccoud, line 96: I suppose you tried this :crossed_fingers:
```suggestion
        from django.core.cache.backends.locmem import LocMemCache
        with mock.patch.object(LocMemCache, "delete", return_value="delete") as locmem_mock:
```

jbpenrath: Yes, `call_count` is still 0 😕

jbpenrath: `locmem_mock.delete.call_count` was 0, `locmem_mock.call_count` is right ...

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,135 @@
...
+        }
+    )
+    @mock.patch.object(DummyCache, "delete")
+    @mock.patch.object(RedisCacheWithFallback, "_call_redis_cache")
+    def test_invalidate_fallback_cache(self, redis_cache_mock, delete_mock):
+        """
+        Test that fallback_cache is invalidated every 60 seconds
+        when Redis is up
+        """
+        client = RedisCacheWithFallback(None, {})
```
sampaccoud, line 99:
```suggestion
        # Getting a value from the redis cache should delete the fallback cache
        client = RedisCacheWithFallback(None, {})
```

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,135 @@
...
+        client.get("irrelevant")
+
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_not_called()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        now = datetime.datetime.now() + datetime.timedelta(seconds=120)
```
sampaccoud, line 107:
```suggestion
        # Another call to Redis cache 120 seconds later should delete the fallback cache again
        now = datetime.datetime.now() + datetime.timedelta(seconds=120)
```

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,135 @@
...
+        with mock.patch("base.utils.datetime") as mock_datetime:
+            mock_datetime.now.return_value = now
+            client.get("irrelevant")
+
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_called_once()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        now = datetime.datetime.now() + datetime.timedelta(seconds=30)
```
sampaccoud, line 117:
```suggestion
        # Another call to Redis cache 30 seconds later (<1 minute) should not delete the fallback cache again
        now = datetime.datetime.now() + datetime.timedelta(seconds=30)
```

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,135 @@
...
+        with mock.patch("base.utils.datetime") as mock_datetime:
+            mock_datetime.now.return_value = now
+            client.get("irrelevant")
+
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_not_called()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        now += datetime.timedelta(seconds=30)
```
sampaccoud, line 127:
```suggestion
        # Another call to Redis cache exactly 1 minute after the first call should not delete the fallback cache again
        now += datetime.timedelta(seconds=30)
```

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,135 @@
...
+        with mock.patch("base.utils.datetime") as mock_datetime:
+            mock_datetime.now.return_value = now
+            client.get("irrelevant")
+
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_not_called()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
```
sampaccoud, line 135: I have 2 suggestions for improvement:
- intialize `now = datetime.now()` only once at the beginning of the test and use another variable `read_time = now + timedelta(seconds=30)`. I know the execution time is negligeable but it seems that this last case you're testing should then be >60 seconds and therefore trigger a new delete?
- get a different key each time to make sure that the throttle is independant of what is retrieved in cache.

jbpenrath: Very good idea, it avoids some big mistakes.

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Test that fallback_cache is invalidated every 60 seconds
+        when Redis is up
+        """
+        now = datetime.datetime.now()
+
+        client = RedisCacheWithFallback(None, {})
+        client.get("round_0")
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
```
jbpenrath, line 104: As `RedisCacheWithFallback` is the default cache in settings and there are also other tests that do cache access, we cannot check cache is invalidated at the first access cache in this test.

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,159 @@
...
+            mock_datetime.now.return_value = read_time
+            client.get("round_4")
+
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_not_called()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        # Another call to Redis cache exactly 1 minute and 1 second after
+        # the first call should delete the fallback cache again
```
sampaccoud, line 150:
```suggestion
        # the latest call should delete the fallback cache again
```

sites/cnfpt/src/backend/base/tests/test_cache.py
```diff
@@ -0,0 +1,159 @@
...
+            mock_datetime.now.return_value = read_time
+            client.get("round_3")
+
+        redis_cache_mock.assert_called_once()
+        delete_mock.assert_not_called()
+        redis_cache_mock.reset_mock()
+        delete_mock.reset_mock()
+
+        # Another call to Redis cache exactly 1 minute after
+        # the first call should not delete the fallback cache again
```
sampaccoud, line 138: Make it easier to understand the test, since it's actually the time elapsed since the latest call and not the first:
```suggestion
        # the latest call should not delete the fallback cache again
```

sampaccoud (approved):
A minor suggestion on comments. 
Congrats for this PR! It will save FUN users numerous down times :clap: :astonished:

PR #197, opened 2021-02-01
Add redirects from openedx routes

sampaccoud:
## Purpose

We don't want to lose all the traffic currently directed to course syllabus on OpenEdX. 

## Proposal
    
Add a view that catches these urls and tries to redirect them to the corresponding course, the corresponding university or as a last resort to the search page.

### DIFF
--- a/sites/funmooc/src/backend/base/views.py
+++ b/sites/funmooc/src/backend/base/views.py
@@ -1,0 +1,40 @@
+"""Views for the fun-mooc site."""
+from django.db.models import Q
+from django.http import HttpResponsePermanentRedirect
+
+from cms.api import Page
+from cms.constants import PUBLISHER_STATE_PENDING
+from richie.apps.core.views.error import error_view_handler
+
+
+def redirect_edx_courses(request, organization, course, session):
+    """
+    The richie site is hosted on the same domain as OpenEdX before.
+    Redirect OpenEdX course urls to the corresponding Richie course urls.
+    """
+
+    def get_redirect_url(**kwargs):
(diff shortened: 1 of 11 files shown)


sites/funmooc/src/backend/base/views.py
```diff
@@ -0,0 +1,41 @@
...
+    Redirect OpenEdX course urls to the corresponding Richie course urls.
+    """
+
+    def get_redirect_url(**kwargs):
+        """Look for a published page matching the kwargs query filters."""
+        try:
+            page = Page.objects.get(
+                ~Q(title_set__publisher_state=PUBLISHER_STATE_PENDING),
+                publisher_is_draft=False,
+                title_set__language=settings.LANGUAGE_CODE,
```
jbpenrath, line 23: Why not use `request.LANGUAGE_CODE`. In this way, we will be ready to enable other languages transparently ?

jbpenrath (approved):
Smart !

sites/funmooc/src/backend/base/views.py
sampaccoud: That's a good idea.

sampaccoud:
@jbpenrath I have added the settings modification to all sites for coherence. If you want to read a last time.

jbpenrath (approved):
Just update ci configuration with `--ignore-changelog` option.

sampaccoud:
> Just update ci configuration with `--ignore-changelog` option.



> Just update ci configuration with `--ignore-changelog` option.

and... it's all green!

PR #205, opened 2021-03-23
✨(funmooc) display first open course run in course detail top header

sampaccoud:
## Purpose

Since we made the video teaser bigger on the course detail page header, the course run enrollment button was pushed below the fold. We want to feature it on top of the page.

## Proposal

Split open course runs in 2: 
- the first one is displayed in the page's header
- the remaining ones are displayed in the side column
- Appropriate notices are displayed when there is only one open course run or when there are no open course runs at all

## Screenshot

image

### DIFF
--- a/sites/funmooc/CHANGELOG.md
+++ b/sites/funmooc/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Display the first open course run in the header of the course detail page
 - Upgrade richie to 2.3.0
 
 ### Added
--- a/sites/funmooc/src/backend/base/tests/test_templates.py
+++ b/sites/funmooc/src/backend/base/tests/test_templates.py
@@ -1,0 +1,122 @@
+"""Test fun-mooc views."""
+import random
+from datetime import timedelta
+
+from django.test import TestCase
+from django.utils import formats, timezone, translation
+
+import lxml.html  # nosec
+from lxml import etree  # nosec
+from richie.apps.courses.factories import CourseFactory, CourseRunFactory
+
+
+class TemplatesTestCase(TestCase):
+    """Test the funmooc template overrides."""
+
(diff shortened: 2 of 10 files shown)


sites/funmooc/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Display the firt open course run in the header of the course detail page
```
jbpenrath, line 13:
```suggestion
- Display the first open course run in the header of the course detail page
```

jbpenrath (approved):
One of your test failed. It seems that `fr/LC_MESSAGES/django.po` committed is outdated.
`lxml` looks pretty cool to ease html content assertions.
Otherwise LGTM ✅

sites/funmooc/src/backend/base/tests/test_templates.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertTrue(page.publish("fr"))
+
+        url = page.get_absolute_url()
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+
+        html = lxml.html.fromstring(response.content)
+
+        # Check syllabus intro
+        header = str(etree.tostring(html.cssselect(".subheader__intro")[0]))
```
sampaccoud, line 65: @carofun just wanted to show you that I used `lxml` without the outdated `django-with-asserts` lib. Maybe a good idea to remove this dependency on https://github.com/openfun/ashley as well.

sampaccoud:
> One of your test failed. It seems that `fr/LC_MESSAGES/django.po` committed is outdated.

In fact, translations were not working on the CI because we forgot to compile them (the .mo files are git ignored).

sites/funmooc/src/backend/base/tests/test_templates.py
carofun: Yes indeed, thank you for the feedback, I keep this example on the side to apply it to our repo.

PR #207, opened 2021-03-24
🐛(styles) import missing stylesheet for LTI consumer plugin

sampaccoud:
## Purpose

The stylesheet for the new LTI consumer plugin was missing resulting in a transparent background for example. 

## Proposal

This import was forgotten because it was not mentionned in upgrade instructions. Add it to the `_main.scss` file.


jbpenrath (approved):
You edited the wrong CHANGELOG (`demo`).

PR #208, opened 2021-03-24
🐛(all) import missing stylesheet for LTI consumer plugin

sampaccoud:
## Purpose

The stylesheet for the new LTI consumer plugin was missing resulting in a transparent background for example. 

## Proposal

This import was forgotten because it was not mentionned in upgrade instructions. Add it to the `_main.scss` file.

This is a follow-up of [this PR](https://github.com/openfun/richie-site-factory/pull/207) that we should have applied on all sites.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -340,94 +340,259 @@ version: 2.1
 workflows:
     cnfpt:
         jobs:
-            - no-change:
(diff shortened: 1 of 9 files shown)


jbpenrath (approved):
This time everything is there.

PR #210, opened 2021-03-25
🚸(search) rename "first session" filter to "new course"

sampaccoud:
## Purpose

Our support team requested to rename the "First session" filter to "New course" after some feedback proved that the wording was not clear.

## Proposal

Rename filter in settings and update translation files.


jbpenrath (approved):
LGTM ✅

PR #215, opened 2021-04-12
Feat/catch edx routes

jbpenrath:
## Purpose
To not loose all the traffic currently directed to OpenEdX, we need to catch OpenEdX course and university routes then redirect them to the corresponding Richie page.

About course, we are already doing that but we did not catch the "old" OpenEdX route (`/<organization_id>/<course_id>/<session_id>`).
Concerning universities, the url to catch is `/universities/<organization_id>`

## Proposal
- [x] Catch old OpenEdX course urls to redirect visitor on the corresponding Richie course page
- [x] Catch OpenEdX university urls to redirect visitor on the corresponding Richie university page

### DIFF
--- a/sites/funmooc/src/backend/base/tests/test_views.py
+++ b/sites/funmooc/src/backend/base/tests/test_views.py
@@ -26,6 +26,24 @@ class CoursesEdxRedirectViewsTestCase(TestCase):
             fetch_redirect_response=True,
         )
 
+    def test_views_redirect_edx_courses_success_with_old_course_uri(self):
+        """OpenEdX course urls are redirected to the corresponding page in richie."""
+        course = CourseFactory(
+            code="abc", page_title="Physique 101", should_publish=True
+        )
+        TitleFactory(page=course.extended_object, language="en", title="Physics 101")
+        course.extended_object.publish("en")
+
+        response = self.client.get("/courses/sorbonne/abc/001/about/")
+
+        self.assertRedirects(
+            response,
+            "/fr/physique-101/",
+            status_code=301,
+            target_status_code=200,
+            fetch_redirect_response=True,
+        )
+
     def test_views_redirect_edx_courses_fallback_organization(self):
         """
         OpenEdX course urls are redirected to the organization page if the course page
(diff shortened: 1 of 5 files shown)


sites/funmooc/src/backend/base/tests/test_views.py
```diff
@@ -49,13 +93,21 @@ def test_views_redirect_edx_courses_fallback_search_page(self):
         nor the organization page can be found.
         """
         PageFactory(
-            reverse_id="search",
+            reverse_id="courses",
```
kernicPanel, line 96: I don't mean there's a mistake here, but I don't get the relation between search and this reverse_id.

sites/funmooc/src/backend/base/tests/test_views.py
```diff
@@ -49,13 +93,21 @@ def test_views_redirect_edx_courses_fallback_search_page(self):
...
         )
 
+        PageFactory(
+            reverse_id="sorbonne",
+            template="richie/single_column.html",
+            title__title="Sorbonne",
+            title__language="fr",
+            should_publish=True,
+        )
+
```
kernicPanel, line 110: I don't get why this factory is needed either…

kernicPanel (approved):

sites/funmooc/src/backend/base/tests/test_views.py
jbpenrath: Because, on fun-mooc.fr, the reverse_id of the search page is `courses` not `search`.

sites/funmooc/src/backend/base/tests/test_views.py
jbpenrath: Just to ensure that if a page has a reverse_id matching organization argument but it is not an organization page, we did not return this page.

Maybe I have to explain this in a comment.

kernicPanel: ok I see, thanks :)

sampaccoud:
@jbpenrath we should not rely on reverse_id as we added a "code" field to hlod this information uniquely.

jbpenrath:
> @jbpenrath we should not rely on reverse_id as we added a "code" field to hlod this information uniquely.

Wow I missed the `organization settings` entry in the Page menu, so I thought code was an autogenerated code.

PR #222, opened 2021-04-28
Funmooc fix contributors filter and rename new courses filter

sampaccoud:
### Fixed

- Fix contributors filter by overriding only its human name
- Rename "New course" option to "New courses" for coherence with filter name


sites/funmooc/src/backend/funmooc/settings.py
```diff
@@ -480,9 +480,9 @@ class Base(StyleguideMixin, DRFMixin, RichieCoursesConfigurationMixin, Configura
                 "is_autocompletable": True,
                 "is_searchable": True,
                 "min_doc_count": 0,
-                "name": "contributors",
+                "name": "persons",
                 "position": 6,
-                "reverse_id": "contributors",
+                "reverse_id": "persons",
```
mbenadda, line 485: If you change the `reverse_id` here you also need to change the reverse ID on the live site, right?

sampaccoud: Yes ! That's the plan if I don't forget :sweat_smile:

mbenadda (approved):
👍

sites/funmooc/src/backend/funmooc/settings.py
mbenadda: Just making sure it was intentional :)

You'll also need to re-generate the ES indices.

PR #225, opened 2021-05-03
📝(readme) fix default dev address in README

mbenadda:
## Purpose

The readme mentions projects running on localhost:8080, while in reality they are running on localhost:8070, as seen in the
docker-compose file.

## Proposal

Change port in the docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -340,314 +340,39 @@ version: 2.1
 workflows:
     cnfpt:
         jobs:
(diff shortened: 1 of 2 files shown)


jbpenrath (approved):
Good catch 🙏 
You have to regenerate the Circle CI `config.yml` with the command `bin/ci check_configuration`

jmaupetit (approved):

PR #243, opened 2021-06-02
🔧(demo) Allow setting demo site domain so it is not localhost:8080

sampaccoud:
## Purpose

On our demo site, the site domain always ended-up as localhost:8080. 

## Proposal

It was because we override it with a default value while building the demo site.

This PR allow setting a custom domain for the site.


jbpenrath (approved):
LGTM ✅

Issue #273, opened 2021-08-09
Setup Django CMS Check SEO plugin


jbpenrath:
A Django CMS plugin to check SEO of each CMS page has been developed, it could be great to check your sites with this tool.

https://github.com/kapt-labs/django-check-seo/

PR #274, opened 2021-08-11
➕ (all) set django-check-seo up

jbpenrath:
# Purpose
 This is a Django CMS Plugin able to scan each CMS pages and generate a reports about SEO performances.

Capture d’écran 2021-08-10 à 23 19 05

Related issue : #273

# Proposal 
- [x] Set Django Check SEO up


sampaccoud:
Maybe add a test that gets the url and minimally checks that it worked?

PR #283, opened 2021-09-01
[ademe] Fix LTI settings and domain for CloudFront proxying for deployment

sampaccoud:
### Purpose

Fix minor issues encountered when deploying the Ademe image to preprod and production environments.

### Proposal

- [x] fix staging and preprod domains for CloudFront proxying
- [x] set LTI providers to marsha video production

While doing these 2 fixes, I noticed that the LTI providers settings had "TEST" in their name which is a copy/pasta fail from Richie's sandbox, so I propose to correct it for all sites.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -343,7 +343,7 @@ workflows:
             - check-changelog:
                 filters:
                     branches:
-                        ignore: /.*/
+                        ignore: master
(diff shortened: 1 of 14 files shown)


jbpenrath (approved):
Good catch ! 
LGTM ✅

PR #284, opened 2021-09-02
Ademe adjustment from feedbacks on 08/31/2021

sveetch:
This is a serie of adjustment as required from customer almost only about layout.

- [x] Use 'burnt-orange' color instead of 'pumpkin-orange'
- [x] Change 'Help' button to 'Contact us'
- [x] Adjust font size on search filters in course list
- [x] Replace TTF fonts with Woff and Woff2 formats
- [x] Fix font-family on all buttons to use 'Gotham'
- [x] Replace contact button on course detail by the current course session (as done on funmooc site)
- [x] Adjust home intro so the new background image can be in full width


sites/ademe/src/backend/templates/richie/base.html
```diff
@@ -5,7 +5,7 @@
 
 {% block topbar_contact %}
 <li class="topbar__item topbar__item--cta">
-    <a href="https://www.fun-services.help/hc/">{% trans "Help" %}</a>
+    <a href="/contact/">{% trans "Contact us" %}</a>
```
sampaccoud, line 8: Since it is a cms page, It's better to do like here https://github.com/openfun/richie-site-factory/blob/master/sites/funcampus/src/backend/templates/richie/base.html#L23 otherwise it may change language or break if the slug changes.

sveetch: nice catch, i was doing it the crappy way..

sampaccoud: another way to say it :stuck_out_tongue:

sampaccoud (approved):

sampaccoud:
@sveetch let's merge everything under one commit mentionning customer feedbacks.

sampaccoud (approved):

PR #293, opened 2021-09-29
Chore/upgrade to richie 2.8.1

jbpenrath:
## Purpose

We want to upgrade all sites to richie v2.8.1

## Proposal

- [x] Run `bin/upgrade 2.8.1 -c` and update CHANGELOG.md
- [x] Upgrade to Django CMS 3.9.0 
- [x] Upgrade frontend dependencies
- [x] Refactor xiti code to build upon https://github.com/openfun/richie/pull/1399

### DIFF
--- a/sites/funmooc/CHANGELOG.md
+++ b/sites/funmooc/CHANGELOG.md
@@ -14,6 +14,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Upgrade richie to 2.8.1
 - Rename `LTI_TEST_*` settings to `LTI_*` as "TEST" does not make sense here
 
 ## [1.8.1] - 2021-06-08
--- a/sites/funmooc/CHANGELOG.md
+++ b/sites/funmooc/CHANGELOG.md
@@ -14,6 +14,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Remove `MARKETING_CONTEXT` and use `WEB_ANALYTICS_*` context instead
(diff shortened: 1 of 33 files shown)


sites/funmooc/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Remove `MARKETING_CONTEXT` and use `WEB_ANALYTICS_*` context instead
```
sampaccoud, line 17: In commit message:
`Richie provides now a web_analytics context > you mean "context variable"

So we do need anymore to the marketing_context > we do not need to add it to `marking_context` anymore

sites/funmooc/src/backend/base/context_processors.py
```diff
@@ -3,48 +3,23 @@
...
-            .distinct()
-            .values_list("code", flat=True)
-        )
-
     def get_privacy_policy_uri():
         """
         Check if a privacy policy page is published in the request language
         then return its url.
         """
+        language = get_language_from_request(request, check_path=True)
```
sampaccoud, line 21: why not keep it outside of this function?

sampaccoud (approved):
Wow amazing :astonished:  :muscle:

sites/funmooc/src/backend/base/context_processors.py
jbpenrath: `language` is now used only in one context so I move this variable near to the context which uses it. 🤷‍♂️

PR #303, opened 2021-12-28
♻️(css) add missing modal styles import and variables

mbenadda:
## Purpose

The upgrade to Richie 2.10.0 required adding an import for modal styles, which were recently integrated in a separate file, to the main stylesheets for all the richie sites.
We also needed to add color entries for "modal", which partly replaces "search-filter-group-modal" and breaks the build for sites which replace the default theme with their own.

This was not documented in upgrade.md and therefore forgotten during the upgrade process.


sampaccoud:
@mbenadda oops this really requires a PR to add an entry in the richie UPGRADE.md file. Maybe give a full diff of what needs to be changed?

sampaccoud (approved):

mbenadda:
Yes, I was planning on updating Richie as well, as soon as I make the other PR and see if there's more to add to `UPGRADE.md` for it.

PR #323, opened 2022-02-23
📈(funmooc) enable xiti course enrollment analytics

jbpenrath:
Since Richie 2.13.0, we are able to track when a user successfully
enrolls to a course. On FUN Mooc, we have to override the frontend part
to add Xiti support and respect user consent.

So from now, when an enrollment request to a course run succeeded, a `click.action` hit with name `courseEnroll` is sent to xiti. This hit contains a custom property `resource_link`. In this way, from Xiti analytic dashboard we are able to segment clicks by resource_link content (so segment by organization, course or course_run)

image

### DIFF
--- a/sites/funmooc/src/frontend/js/utils/api/web-analytics/index.ts
+++ b/sites/funmooc/src/frontend/js/utils/api/web-analytics/index.ts
@@ -1,0 +1,18 @@
+import type { Maybe } from 'richie-education/js/types/utils';
+import { WebAnalyticsAPI } from 'richie-education/js/types/WebAnalytics';
+import context from 'richie-education/js/utils/context';
+import { handle } from 'richie-education/js/utils/errors/handle';
+import FunXitiApi from '../../../../../../js/utils/api/web-analytics/xiti';
+
+const WebAnalyticsAPIHandler = (): Maybe<WebAnalyticsAPI> => {
+  try {
+    if (context?.web_analytics_provider === 'xiti') {
+      return new FunXitiApi();
+    }
+  } catch (error) {
+    handle(error);
+  }
+  return undefined;
+};
+
+export default WebAnalyticsAPIHandler;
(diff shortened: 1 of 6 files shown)


sites/funmooc/src/frontend/js/utils/api/web-analytics/index.ts
```diff
@@ -0,0 +1,31 @@
...
+      case WebAnalyticsAPIBackend.GOOGLE_TAG_MANAGER:
+        return new GoogleTagManagerApi();
+      case FunMoocWebAnalyticsAPIBackend.XITI:
+        return new FunXitiApi();
+    }
+  } catch (error) {
+    handle(error);
+  }
+  return undefined;
+};
```
sampaccoud, line 29: I find this weird. Since you are setting `FunMoocWebAnalyticsAPIBackend` to XITI, why not directly return FunXitiApi()?

sites/funmooc/src/frontend/js/utils/api/web-analytics/xiti.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable class-methods-use-this */
+import { BaseWebAnalyticsApi } from 'richie-education/js/utils/api/web-analytics/base';
+
+/**
+ *
+ * Xiti Richie Web Analytics API Implementation
+ *
+ * This implementation is used when web analytics is configured has `xiti`.
```
sampaccoud, line 8:
```suggestion
 * This implementation is used when web analytics is configured as `xiti`.
```

sampaccoud (approved):

sites/funmooc/src/frontend/js/utils/api/web-analytics/index.ts
jbpenrath: You are right. I'll change it.

PR #331, opened 2022-03-30
⬆️(all) upgrade black to version 22.3.0

jbpenrath:
A click update broke black. This has been fixed in version 22.3.0.
psf/black#2964

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -344,7 +344,7 @@ workflows:
             - check-changelog:
                 filters:
                     branches:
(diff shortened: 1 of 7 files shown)


kernicPanel (commented):
Maybe explain why circleci config had to be updated ?

jbpenrath:
> Maybe explain why circleci config had to be updated ?

This is a recurrent task specific to this repository.
The configuration file is generated to trigger jobs only for site which have been modified.

